PR #35, opened 2018-09-27
Revert "Add Travis to CI"

Alan-Cha:
Reverts strongloop/oasgraph#34


Alan-Cha:
I don't think the Travis tests are working. The problem is that `npm run test` needs the example api running, e.g. (`npm run api`), to work. I don't know enough about Travis in order to change that.

Alan-Cha:
While continuous integration is something we want to have in the future, we will need to look into it on a later date because of our particularly complex testing. We need to run multiple processes to run the API, the GraphQL API, and the test queries and we do not have the time to look into how to do so as of now. 

We currently use a system that automatically runs test cases after every commit which may be adequate for R&D and less so in an open source project, they are sufficient for the time being.

Issue #39, opened 2018-10-02
Change our testing system


Alan-Cha:
We currently use the [pre-commit](https://www.npmjs.com/package/pre-commit) NPM module to do our testing. Before every commit, a series of tests will run and if any of the tests fail, then the module will cancel the commit. This restriction can be detrimental to development in certain cases and the testing data is not public. 

We would like change to [Travis CI](https://travis-ci.org). We have tried to do so with pull request #34, however it was unsuccessful because our testing environment is somewhat involved and the changes made did not satisfy the necessary requirements. We need to run multiple processes to run the example API, the graphQL API, and the test queries.

ErikWittern:
I will look into spinning up the API as part of running `npm test`.

PR #41, opened 2018-10-04
Feat cli add remote file capability

marioestradarosa:
This PR adds the functionality for the CLI tool to provide a remote file name as a source for a SOA specification content and an update to the documentation to support the new feature. 

**Features**
- local or remote files 
- ability to specify a port number, which is good if you want to run multiple GraphQL severs at the same time
- documentation updated

**Check List**
[ ] Passed all tests. No. Please check there is a problem with one of the tests not related to this work
[x] Updated documentation


marioestradarosa:
@Alan-Cha, IMO, the new features for CLI doesn't affect any of these tests, as there no tests for the CLI itself 😄 nor it is used in any of the tests directly right?

- test/example_api.test.js
- test/authentication.test.js

I could see that it tries to connect to a remote API server, shouldn't be better if you mock it up in your tests ?

Alan-Cha:
@marioestradarosa Thank you for your contribution! The port number feature in particular is fantastic! @ErikWittern and I will review the code changes. I think there some small formatting issues that we would like to address. After that is done, would you also squash your commits? Sorry for nitpicking but I hope you understand! We are also trying to update the [CONTRIBUTING.md](https://github.com/strongloop/oasgraph/blob/master/CONTRIBUTING.md) to clearly state our expectations. 

Regarding testing, yes, the CLI tool should not affect our tests. To start the example API, run `npm run api`. Then, in another process, run `npm run test` to test. We should make `npm run test` start the API as well and that is related to issue #39, however, personally I'm not too sure how to do it because I'm not great at shell scripts 😅

cli.js
```diff
@@ -9,46 +9,77 @@ const express = require('express')
...
+let filePath = process.argv[2];
+let portNumber = process.argv[3] ? process.argv[3] : 3001;
+
+// check if the file exists 
+if (fs.existsSync(path.resolve(filePath))) {
+      let oas = JSON.stringify(require(path.resolve(filePath)));
+      startGraphQLServer(oas, portNumber);
+    
+} else { // falls back to a remote location
+    if (filePath.includes('http')) {
```
ErikWittern: Better use something like `/^https?/.test(filePath)` here, because it might just be that `'http'` appears in the middle of a filepath.

docs/tutorials/quickstart.md
```diff
@@ -22,31 +22,36 @@ To install Node.js, click [here](https://nodejs.org/en/download/).
...
+```sh
+
 git clone git@github.com:strongloop/oasgraph.git
 cd oasgraph
 npm link
+
 ```
 
-### Save OAS
+### Locate the OAS
```
Alan-Cha: I think we can clarify this by saying the user can either save or locate the OAS. I think mentioning one and not the other may be confusing.

cli.js
```diff
@@ -9,46 +9,77 @@ const express = require('express')
...
+      } 
+}
+
+/**
+ * reads a remote file content using http protocol
+ * @param {string} url specifies a valid URL path including the port number
+ */
+function getRemoteFileSpec (url) { 
+  return new Promise(resolve => { 
+    http.get(url, (resp) => {
```
Alan-Cha: The built-in `http` module does not support https. 

Because we already use the `request` module, @ErikWittern proposed we replace the code with the following:

```
function getRemoteFileSpec (uri) {
  return new Promise((resolve, reject) => {
    request({
      uri,
      json: true
    }, (err, res, body) => {
      if (err) {
        reject(err)
      } else if (res.statusCode !== 200) {
        reject(new Error(`Error: ${JSON.stringify(body)}`))
      } else {
        resolve(body)
      }
    })
  })
}
```

docs/tutorials/quickstart.md
Alan-Cha: ### Save or locate the OAS

OASGraph relies on the OpenAPI Specification (OAS) of an existing API to create a GraphQL interface around that API. OASGraph can also retrieve a web-hosted OAS. 

If you are using LoopBack, you can simply copy the URL location of the web-hosted OAS, which is usually [http://127.0.0.1:3000/openapi.json](http://127.0.0.1:3000/openapi.json).

***

If you want to generate a GraphQL interface for another API, make sure that API is running and proceed in the same way or save the OAS.

Alan-Cha:
Hi @marioestradarosa, thanks once again for the pull request! We added line comments into the commits. Here some general points:

* The current implementation does not work with HTTPS. As we already use the [`request`](https://www.npmjs.com/package/request) library, please switch the implementation to use that, as it will handle these cases. We provide an example in one of the specific comments.
* We added one comment about detecting URLs (vs. file paths) that you might want to consider. We also commented on the updated getting started guide.
* We do so far have no CLI tests for now, but checked your changes and things seem to work well.
* Our code formatting adheres to the [standard.js style](https://standardjs.com/). We yet have to setup automatic checks / fixes for that. Until then, we rely on manual efforts to adhere to the style. Specifically, please ensure the use of two whitespaces (and multiples thereof) as well as avoid semicolons.
* After making the above changes, please squash all your commits using `git rebase -i HEAD~[number of commits]` - we want pull requests to ultimately consist of a single commit.

docs/tutorials/quickstart.md
```diff
@@ -22,31 +22,36 @@ To install Node.js, click [here](https://nodejs.org/en/download/).
...
 
 OASGraph relies on the OpenAPI Specification (OAS) of an existing API to create a GraphQL interface around that API.
 
-If you use LoopBack, start your API server in the LoopBack project folder, and access the API's OAS at [http://127.0.0.1:3000/openapi.json](http://127.0.0.1:3000/openapi.json). Save the OAS to disk.
+If you use LoopBack, start your API server in the LoopBack project folder. Make sure you copy the url location for the OAS specification file which is usually located at [http://127.0.0.1:3000/openapi.json](http://127.0.0.1:3000/openapi.json).
 
-```
-# in the LoopBack project folder:
+#### in the LoopBack project folder
```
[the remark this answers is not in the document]
marioestradarosa: I am using VsCode and it was directing me to this URL for this 
https://github.com/DavidAnson/markdownlint/blob/v0.11.0/doc/Rules.md#md025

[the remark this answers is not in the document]
marioestradarosa: But I will rever it back.

marioestradarosa:
I wonder why [cis-jenkins] is failing, while the local tests all passed.

screen shot 2018-10-04 at 13 37 11

marioestradarosa:
>  we want pull requests to ultimately consist of a single commit.

That won't be necessary good in all cases. If you keep a linear history in the commits:

- it doesn't matter how many commits are per request, they will be adding to the commit history even if they come from different pull requests
- If you enforce it, you will see the effects later on. An example would be to pin point a bug introduced in a large PR that was forced to squashed into 1 single commit. 
- Also it is good to have  a commit doing a specific task that can be reviewed easily.

marioestradarosa:
> We assumed this was a rule for all repositories in strongloop, but that seems not to be the case

I think it is really a criteria to keep the commit history clean for later revision and not abuse of commits.

But for example, imagine the case you add a feature along with its documentation.  In this case during reviewing we can push **_N_** number of _commits_ but when it is _ready_ to merge, we could easily squash the feature, test and its documentation in 2 separate commits regardless of the number of commits before merging, more or less 😄 , at the end it is your call.

PR #42, opened 2018-10-04
Programmatically start and stop example API during tests to address #39

ErikWittern:
Signed-off-by: Erik Wittern <witternj@us.ibm.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,17 +194,13 @@ Because OASGraph is a library, it cannot make the callbacks that OAuth requires 
 To see an example of how this would work, click [here](https://github.ibm.com/apiharmony/oasgraph-oauth-github-example)!
 
 ## Testing
-To test OASGraph, first make sure the example API server is running:
-
-```bash
-npm run api
-```
-
-Then, run tests:
+To test OASGraph, run:
 
 ```bash
 npm test
 ```
+
+This command will temporarily start and later shut down an example REST(-like) API.
 
 ### Research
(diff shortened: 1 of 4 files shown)


ErikWittern:
@Alan-Cha To test this, clone the repo, run `npm i`, and then you should be able to just do `npm test` to run all tests.

Alan-Cha:
The changes you made seem to work but I have some concerns. 

First, the tests output the API calls. I think this is due to the fact that the processes' IOs are somehow intertwined. Instead of seeing "pass... pass... pass...", `authentication.test.js` and `example_api.test.js` have long stretches of output data in between. 

Secondly, could you should squash your commits?

marioestradarosa:
> Secondly, could you should squash your commits?

@ErikWittern , in this case it makes sense, since we would like to keep a clear linear history any merge to master should be squashed. My recommendation to keep merge to masters away from the commit history in the PRs is  (albeit there are multiple ways, this works for me):

### To update your local master branch

```sh
git checkout master
git fetch upstream
git pull upstream master
```

### To update your working branch
```sh
git checkout <yourbranch>
git rebase master
git push --force-with-lease
```

ErikWittern:
@Alan-Cha The first issue has to do with the example API server and the tests running in the same process now. I.e., the console logs of the example API server are printed between test results. Not perfect indeed, but I would argue also not a show stopper for going forward with this pull request. In the future, we can consider making logs of the example API server mutable.

ErikWittern:
@marioestradarosa Thanks for your advise! I think I managed to squash the two commits that way, effectively removing the "Merge" commit.

PR #43, opened 2018-10-05
feat(cli) add option to save schema

marioestradarosa:
Signed-off-by: Mario Estrada <marioestradarosa@yahoo.com>

This PR adds the ability to save a generated **GraphQL** **Schema** to a file. The generated file is for now `schema.graphql`. The PR was built as proposed by @ErikWittern https://github.com/strongloop/oasgraph/issues/16.

**Added:** 
- the port number and --save boolean switch are mutually exclusive 
- specifying --save does not start the server, only generates the file and informs user


cli.js
```diff
@@ -10,18 +10,33 @@ const graphqlHTTP = require('express-graphql')
...
+let portNumber = 3001
+let saveSchema = false
+
+// examine the last two arguments
+for (var i=3; i<=process.argv.length-1; i++) {
+   if (isNaN(process.argv[i]) && process.argv[i] === '--save') {
+     saveSchema = true
+   } else {
+     portNumber = process.argv[i]
+     }
```
Alan-Cha: Could you remove this extra space?

cli.js
```diff
@@ -10,18 +10,33 @@ const graphqlHTTP = require('express-graphql')
...
 
 let filePath = process.argv[2]
-let portNumber = process.argv[3] ? process.argv[3] : 3001
+
+// default values if only the url or local file is specified
+let portNumber = 3001
+let saveSchema = false
+
+// examine the last two arguments
+for (var i=3; i<=process.argv.length-1; i++) {
```
Alan-Cha: This may be a problem if a user adds more arguments then intended. It might be okay for now but could be hazardous if we want to add more options, refer to #44, What do you think @ErikWittern?

marioestradarosa: I agree. I believe in a near future, we should integrate something like  https://www.npmjs.com/package/minimist to handle more complex command line options. I will place a MAX_PARAM constant for now. 
thoughts?

Alan-Cha: Yes, I completely agree. Setting up the CLI interface from scratch will definitely lead to trouble. We should look into it if we want to build out the CLI tool more.

marioestradarosa: ok. for now, if you inspect the code, the `--save` option is safe and now for port number the user can specify multiple argumetns if he wants, it will only take the ones that are numbers, so the user can enter:

`oasgraph openapi.json 3002 --save mickey pluto`

and still will work , and if user enters 

`oasgraph openapi.json 3002 --save mickey pluto 3003`

it will start on port number 3003 , any way, I think it is safe now 😄

cli.js
marioestradarosa: please check if I understood or specify clearly what extra space you meant?

Alan-Cha: Yes, this is perfect. Thank you for making that change for me!

PR #45, opened 2018-10-05
Set up Travis CI

Alan-Cha:
The badge will not work in preview because it's pointed at the Strongloop branch but the following badge should work because it's pointed at my local repository. 

Travis (.org)

The next step is to decide what to do about pre-commit, should we just get rid of it?


Alan-Cha:
Related to issue #39

PR #48, opened 2018-10-05
Set up Travis CI

Alan-Cha:
Added tutorial video
Small change to tutorial

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+sudo: false
+language: node_js
+node_js:
+  - '8'
+  - '10'
+
+os:
+  - linux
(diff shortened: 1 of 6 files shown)


Alan-Cha:
Check out that Travis CI check! I wish we can get rid of those jenkins test though, will look into it.

virkt25:
engine node should be added so end users of the package know what version the package is compatible with. This doesn't affect Travis builds **BUT** this will impact what versions of Node Jenkins will test against. 

To disable Jenkins you need to do so in the internal repo that does controls it. See: apimesh/ci in internal GitHub.

PR #51, opened 2018-10-08
Fixed some problems with the documentation

Alan-Cha:


marioestradarosa:
@ErikWittern , with this commit, the the `--save` documentation was removed which was previously commited, was this intentionally, you will remove the capability of saving the generated schema ?

Alan-Cha:
@marioestradarosa Sorry for the misunderstanding. We simply moved the documentation for the `--save` feature into the README.md. 

Actually, this is what I thought you did in the pull request to begin with but I should have looked more closely. 

I think we want to keep the quick start guide as simple as possible, the minimum information needed to start using OASGraph. While I think the `--save` feature is important, I do not think it was necessary to keep it in the quick start guide so I put it in the usage section of the README.md. I hope you understand!

PR #74, opened 2018-11-20
Fixed #18

### LINKED ISSUE #18: createOrReuseDataDef() can create inappropriate object type names for simple schemas
OASGraph reuses object type names based on schemas. This can be a problem if an OAS contains a number of simple schemas, i.e. only have a single property, because these schemas will be associated with the same object type name even though they are used in different contexts. 

A good example of a simple schema would be the following: 

```
{
     "type": "string"
}
```

If this schema was used to describe both a Dog and a Cat object, then they may be both given the same object type name, e.g. both "Dog" or "Cat", by createOrReuseDataDef().

***

This problem shouldn't be hard to fix. Just write an exception for this edge case. The function getSchemaType() could be helpful. 

***
How I discovered this problem:

Because our test cases mainly deal with complex object types, we did not experience many collisions. Also, this issue will not create errors so it may have just slipped through the cracks. 

I encountered when I was trying to enhance non-application/json capabilities. I had to treat the request and response bodies as black boxes. In order to do so, I changed the schemas into the simple schema described above. Because they all had the same schema, they ended up receiving the same object type name. Only when I changed the **description** did they receive the expected object type names.

Alan-Cha:
Using minimal changes. Should give function createOrReuseDataDef() a parameter, preferredName, for lookup.


packages/oasgraph/src/preprocessor.ts
```diff
@@ -334,6 +336,7 @@ export function createOrReuseDataDef (
 
   let def = {
     schema,
+    preferredName: preferredName,
```
ErikWittern: You can simply do `preferredName,` here.

Issue #56, opened 2018-10-10
Improve CLI and provide tests


ErikWittern:
The CLI currently depends on custom logic to parse arguments. As we expand the CLI's capabilities, we should consider using a framework like [commander.js](https://tj.github.io/commander.js/). This is related to previous issue #44.

We should also refactor the code into a "core" and a "arguments parsing" part to be able to write test cases against the core part.

Alan-Cha:
We currently put the [commander.js](https://tj.github.io/commander.js/) library to good use but unfortunately, we still don't have a testing framework set up for the CLI. I think this is a much needed improvement.

wtrocki:
From my experience it best to not do any processing in CLi handler and just delegate to serparate method that can be tested outside using ava or jest. Testing cli it is just going to test commander.js or yargs which makes this process more complex.

Alan-Cha:
@wtrocki Thanks for the suggestion. I see what you mean. The CLI is just a thin wrapper for OtG so do you think it makes sense to have a testing framework for it?

wtrocki:
I'm really lacking context where this is used but cli creates the fully functional server that will give some nice integration testing capabilities. Example of such tests here:
https://github.com/aerogear/voyager-server/tree/master/examples/test

PR #77, opened 2018-11-21
Integrated Commander js into CLI

Alan-Cha:
Related to issue #56


packages/oasgraph-cli/README.md
```diff
@@ -19,29 +19,49 @@ npm i -g oasgraph-cli
...
 
 ## Usage
-The basic usage of the CLI takes the specified OAS, create a GraphQL interface for it, and start a server to host the GraphQL interface.
 
-You can specify a local file containing the OAS specification or a remote url such as `http://127.0.0.1:3000/openapi.json`. Additionally, you can specify an optional port number so you can have multiple GraphQL servers for testing in the same machine.
+```
+Usage: oasgraph <OAS JSON file path or remote url> [options]
+
+Options:
+  -V, --version             output the version number
```
ErikWittern: I think this is a bit unspecific - both the documentation and the display of the version. Are we talking the version of OASGraph or the CLI here? I propose to output both, and make that clear in the docs also.

packages/oasgraph-cli/README.md
```diff
@@ -19,29 +19,49 @@ npm i -g oasgraph-cli
...
+  -p, --port <port>         select the port where the server will start
+  -s, --strict              create a GraphQL schema matching the given OAS perfectly
```
ErikWittern: I am not sure about the wording: wouldn't the user expect a "perfectly" matched schema? Maybe we need a little more explanation here.

packages/oasgraph-cli/README.md
```diff
@@ -19,29 +19,49 @@ npm i -g oasgraph-cli
...
+  -a, --addSubOperations    nest operations based on path hierarchy
+  -f, --fillEmptyResponses  wrap specific operations with no response schema
```
ErikWittern: Wrap with what? Also, maybe instead of "specific", we could point to the 204 status code, which I believe is a mandatory precondition?

packages/oasgraph-cli/README.md
```diff
@@ -19,29 +19,49 @@ npm i -g oasgraph-cli
...
+  --no-viewer               create artificial constructs that pass authentication credentials
```
ErikWittern: Maybe just: `Create viewer object types for passing authentication credentials.`?

packages/oasgraph-cli/README.md
```diff
@@ -19,29 +19,49 @@ npm i -g oasgraph-cli
...
+  --save <file path>        save schema to path
```
ErikWittern: Is this instead of starting the GraphQL server?

packages/oasgraph-cli/README.md
Alan-Cha: The `-V` and `--version` options are auto-generated by Commander however I will see if I can overwrite them.

packages/oasgraph-cli/README.md
Alan-Cha: Would something along the lines of `throw an error if OASGraph cannot create a GraphQL schema matching the given OAS perfectly` be better? 

Here is the [options doc](https://github.com/strongloop/oasgraph/blob/master/packages/oasgraph/README.md#options)

packages/oasgraph-cli/README.md
Alan-Cha: Wrap with GraphQL? Maybe `run OASGraph on legitimate operations with no response schema, such as those with HTTP 204 code, rather than ignore them`?

Yes, "specific" refers to 204 but I wasn't sure if we wanted to expand the preconditions.

packages/oasgraph-cli/README.md
Alan-Cha: Yes, I will add that then.

packages/oasgraph-cli/README.md
Alan-Cha: Unfortunately, I don't think this can be changed easily

```
alans-mbp:oasgraph-cli alan.cha$ node lib/oasgraph.js -h
Usage: oasgraph <OAS JSON file path or remote url> [options]

Options:
  -V, --version             output the version number
  -V, --version             output the version number of the OASGraph library and CLI
  -p, --port <port>         select the port where the server will start
  -s, --strict              create a GraphQL schema matching the given OAS perfectly
  -a, --addSubOperations    nest operations based on path hierarchy
  -f, --fillEmptyResponses  wrap specific operations with no response schema
  --no-viewer               create artificial constructs that pass authentication credentials
  --save <file path>        save schema to path
  -h, --help                output usage information
```

It trying to change the version makes the overall code very ugly

packages/oasgraph-cli/README.md
ErikWittern: I like it better

packages/oasgraph-cli/README.md
ErikWittern: `create placeholder schema for operations with HTTP status code 204 (no response) rather than ignore them`?

packages/oasgraph-cli/README.md
ErikWittern: Ok, then let's stick with the original

packages/oasgraph-cli/README.md
```diff
@@ -19,29 +19,49 @@ npm i -g oasgraph-cli
...
-You can specify a local file containing the OAS specification or a remote url such as `http://127.0.0.1:3000/openapi.json`. Additionally, you can specify an optional port number so you can have multiple GraphQL servers for testing in the same machine.
+```
+Usage: oasgraph <OAS JSON file path or remote url> [options]
+
+Options:
+  -V, --version             output the version number
+  -p, --port <port>         select the port where the server will start
+  -s, --strict              throw an error if OASGraph cannot run without compensating for errors or missing data in the OAS
+  -a, --addSubOperations    nest operations based on path hierarchy
+  -f, --fillEmptyResponses  run OASGraph on operations with no response schema, such as those with HTTP 204 code, rather than ignore them
```
ErikWittern: I propose: `create placeholder object type for operations with HTTP status code 204 (no response) rather than ignore them`.

PR #84, opened 2018-12-05
Added YAML support for the CLI

Alan-Cha:
I know the nested try catch is pretty ugly but not sure how to make it any prettier...


ErikWittern:
Again, nice work, @Alan-Cha !

Issue #96, opened 2019-01-28
Allow to select which scheme to use


thejibz:
When different schemes are availaible (http/https), it would be nice to select which one to use.
I will look into it, and maybe propose a PR.

Alan-Cha:
I’m not too familiar with this kind of usage. Could you clarify or even provide an example?

In any case, we would be ecstatic to see people contribute to our project! That’s the whole point of open sourcing, of course. Please let me know if you need any help! I can help walk you through the code (it’s a bit complex, I have to be honest) or give you suggestions once I have a better idea of what you mean. We really want to make this as easy as possible for people to contribute!

thejibz:
Hello, thank you for your warm welcome !
I was talking of transport schemes available for servers:

- In OAS v2: https://swagger.io/docs/specification/2-0/api-host-and-base-path/ (section "schemes")

- In OAS v3: https://swagger.io/docs/specification/api-host-and-base-path/ (section "Server URL Format")

Simply put, if I can hit the API with either HTTP or HTTPS or WS, how can I tell OASgraph to choose my prefered way (instead of currently taking the first server of the list ) ?

PR coming soon :)

Alan-Cha:
Wow! Thank you so much! This was so quick too! I left a few comments for you to take a look at. Tell me what you think!

Alan-Cha:
This is a cool change because we’ve always randomly picked a base URL. This provides some level of control. Do you think it’s worthwhile to add more fine-grained options to control the base URL? Or do you think it makes more sense just to change the OAS? I was wondering if you, as the pioneer of this idea, have any insight into this.

thejibz:
> Wow! Thank you so much! This was so quick too! I left a few comments for you to take a look at. Tell me what you think!

It's my turn to congratulate you for your code base. It's fairly easy to get into it ! Well structured and commented. Appreciated working on it :+1:

thejibz:
> This is a cool change because we’ve always randomly picked a base URL. This provides some level of control. Do you think it’s worthwhile to add more fine-grained options to control the base URL? Or do you think it makes more sense just to change the OAS? I was wondering if you, as the pioneer of this idea, have any insight into this.

Humm, in my current use case it suffices to let me pick the protocol I prefer.  

But in an ideal world we should have the means to make a round robin on a sublist of OAS' defined servers.
I could see it as a list of servers injected in the gql context and then the resolvers would use one of it, following a policy.
[...]

PR #97, opened 2019-01-28
Add preferedScheme option

thejibz:
Issue #96


packages/oasgraph/src/oas_3_tools.ts
```diff
@@ -194,36 +195,56 @@ export function getBaseUrl (
...
 
   throw new Error('Cannot find a server to call')
 }
 
+/**
+ * Return the first server that use the prefered scheme. If none match, return the first server.
+ * @param servers List of servers
+ * @param preferedScheme Optional. Prefered scheme to use
+ */
+function getPreferedServer(servers: ServerObject[], preferedScheme?: Schemes) {
```
Alan-Cha: I think this code can be expressed in a more intuitive way. What do you think about the following? 

```
function getPreferredServer(servers: ServerObject[], preferredScheme?: Schemes) {
  if (Array.isArray(servers)) {
    if (servers.length === 0) {
      // No choice in servers can be made
      return servers[0]
    } else if (servers.length >= 1) {
      let preferredServers = servers.filter(server => server.url && server.url.startsWith(`${preferredScheme}:`))
        
      if (preferredServers.length > 0) {
        if (preferredServers.length > 1) {
          logHttp(`Warning: Randomly selected the first server that match the preferredScheme option (${preferredScheme}): ${preferredServer.url}`)
        }

        return preferredServers[0]
      } else {
        // No preferred scheme could be found
        logHttp(`Warning: Randomly selected first server ${preferredServer.url}`)
        return servers[0]
      }
    }
  }
  throw new Error('Cannot find a server to call')
}
```

ErikWittern: In the above code, `if (servers.length === 0) {` (line 3) is wrong. The if statement tries to cover the case when there is only a single server, in which case `server.length` should be `1`, though, not `0`.

thejibz:
> A few more comments... we appreciate your efforts to help make our style more consistent.
> 
> I think adding spaces within curly brackets, removing spaces before colons, and tabulating string concatentations is good. However, would it be possible to retain the space before parameters in function signature definitions?
> 
> For example: we would prefer `function sanitize (str: string): string` over `function sanitize(str: string): string`.
> 
> Also we’ve always used curly brackets for `if` and `else` statements.
> 
> Again, the fixes are welcomed but I think this is one of our better followed conventions.
> 
> In addition, would you be able to write some test cases? This is a bit tricky because this change cannot be tested in the same fashion as our usual Example API tests. If you’re looking for suggestions, I’m not sure what would be the best way to test for this but I think @erik may have some ideas. Perhaps we just need to test the function by itself rather than in the context of a working GraphQL interface.
> 
> Again, thanks a lot! Sorry for nitpicking!

No problem,  it's on me to adopt your code style ! I will change my formatter's settings in the context of your project. Also, I agree on the rule "always use curly brackets in if statements".

Regarding, the tests I didn't make one because of the reason you evoke. Not easy to do, but a unit testing on just the function will do the trick I think.

thejibz:
I had to export the getPreferredServer function in order to test it though. Testing the n+1 getBaseUrl would be much harder.

thejibz:
Hello guys !
I had to publish a npm package with the name "oasgraph-jibz" since the "git+https://..." style dependency don't work with the repo pattern root/package/oasgraph :/

I will delete this package as soon as you publish a new version with either preferredScheme or baseUrl option.

PR #100, opened 2019-01-29
Update dependencies

Alan-Cha:


ErikWittern:
I don't see any problems with this change, and the tests went through, so I accept.

Thank you, @Alan-Cha !

PR #105, opened 2019-02-05
Added support for constant and variable link params

Alan-Cha:


ErikWittern:
@Alan-Cha Hah, this is fantastic, it expands the capabilities of what OASGraph can do. Thanks a lot!

PR #106, opened 2019-02-06
Added description for baseUrl and cors

Alan-Cha:


ErikWittern:
@Alan-Cha Thanks for complementing the docs.

Issue #108, opened 2019-02-11
Missing Query Schemas


simax:
I recently reported in [issue #92](https://github.com/strongloop/oasgraph/issues/92) that I now no longer produce an ```OAS``` in ```Swagger 2.0``` format but now produce one  in ```OpenApi 3.0.1``` format. : 

[swagger.txt](https://github.com/strongloop/oasgraph/files/2850813/swagger.txt)

Even though I seem to get better results with that, the results are still somewhat disappointing when using ```OASGraph``` to generate the schema and trying to use the generated ```OAS``` in ```GraphiQL```.

The following problems still occur: 

1. SearchOrders (and possibly others) is missing from the schema and docs even though it's in the Swagger.json file. 

2. Also, as mentioned in [issue #107](https://github.com/strongloop/oasgraph/issues/107) how can I get the docs to present the schema in alphabetical order?

Alan-Cha:
Hi Simon! Thank you for your persistence in trying to make OASGraph work! Your feedback is so valuable to us as it helps us identify what we need to do to make OASGraph better. I hope you will continue to stick with us as we try to address these issues. 

The missing operation is very strange. I briefly looked into it with @ErikWittern and I am about to take a deep dive into it right now. 

I addressed #107 in the latest commit! Everything should be nice and alphabetized as soon as the pull request goes through.

simax:
Cool, thanks guys, the alphabetical order fix is working nicely!

Alan-Cha:
I think I know what the problem is! 

In GraphQL, field names must be unique. In the OAS you provided, there are a few operations that will receive the same field name from OASGraph. They are the following:

* `/api/v1/settings/plans/{id}` and `/api/v1/settings/plans/current`, which create a `plan` field
* `/api/v1/orders` and `/api/v1/orders/search`, which create an `orders` field
* `/api/v1/categories`and `/api/v1/categories/search`, which create a `categories` field

***

I think it's bad that we don't throw a warning so I'll try to change that. Aside from that, it's hard to say what's the best way to move forward...

We could try to put some automatic mitigation in place... For example, we could append a number at the end of a duplicated field name. However, this can create other types of name collisions. I can go into more detail if you'd like. 

Another method would be to define a new schema just so you can provide another name but this is a pretty ugly solution...

What do you think?

ErikWittern:
@Alan-Cha In such cases, if present, how about relying on the `operationId`?

Alan-Cha:
[...]

PR #111, opened 2019-02-13
Added warning for duplicate field names

Alan-Cha:
Related to #108


ErikWittern:
@Alan-Cha Looks good to me. The warning is a first important step to address the issue raised in #108. Thank you!

PR #113, opened 2019-02-26
Closes #38 by providing a definition file

### LINKED ISSUE #38: Publishing source files onto NPM
Currently, we are publishing the source files, specifically the Typescript files under `src/`, in our NPM module. Perhaps this is not the best strategy.

treffynnon:
Adds a definition file for the oasgraph API. It cannot use /types/Options.ts as this has some fields marked as mandatory and they are not.

I am not positive how the npm package is built for this project so this might not be in strictly the right place, but hopefully it is of some use or a basis for further discussion on this issue.

### DIFF
--- a/packages/oasgraph/index.d.ts
+++ b/packages/oasgraph/index.d.ts
@@ -1,0 +1,77 @@
+declare module 'oasgraph' {
+	import { Oas2 } from 'src/types/oas2'
+	import { Oas3 } from 'src/types/oas3'
+	import { Report } from 'src/types/options'
+	interface IOptionsWithOptionals {
+		/**
+		 * Adhere to the OAS as closely as possible. If set to true, any deviation
+		 * from the OAS will lead OASGraph to throw.
+		 */
+		strict?: boolean
+		/**
+		 * Custom headers to send with every request made by a resolve function.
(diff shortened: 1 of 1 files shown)


Alan-Cha:
@treffynnon Hey, thank you for this contribution! This is fantastic. It will definitely improve the lives of TypeScript developers using OASGraph.

However, after discussing it with my colleague @ErikWittern, we believe the best way to move forward is to auto-generate the definition file using the Typescript `declaration` compiler option. We are also going address issues with the `Options` type, as you have already started. 

Please do not take this the wrong way! We will not accept this pull request but your contribution is highly appreciated! We are simply going to take a slightly different method of addressing these issues. Thank you once again and keep an eye out for the next commit!

treffynnon:
That's great news. I just wanted to prompt discussion and generating a
declaration is far better. Please tag a new release with the definitions
soon :)

Alan-Cha:
@treffynnon OASGraph v0.14.1, which has Typescript declaration files, is official out! 🎉

treffynnon:
Perfect! I've deleted my custom types/declarations and using the package supplied one already. Thank you.

PR #126, opened 2019-03-11
Fixed #122

### LINKED ISSUE #122: No test for viewerAnyAuth
Somehow we don't have any tests for viewerAnyAuth. There is only one that counts the number of operations that can be called within the viewer.

Alan-Cha:
* Added another test for operationIdFieldName
* Fixed AnyAuth
* Added tests

There are inconsistencies with how we handle data (where and when we beautify) and I tried to make it more consistent.


Alan-Cha:
I also included a test in which two operations with different auth schemas are bound by a link. Hopeful for links between different OASs.

ErikWittern:
@Alan-Cha Haha, good to have a test for the AnyAuthViewer now... Great work!

Issue #128, opened 2019-03-14
Links with operationRef needs operationId


Alan-Cha:
We currently support links that use `operationRef` but with one caveat, the operation must also have an `operationId`, which according to the specification is optional. Ideally, we can support links with `operationRef` but without the need of requiring `operationId`.

Alan-Cha:
OASGraph generates `operationId`s if they are missing. This function was refactored in #129.

Issue #82, opened 2018-11-29
How to generate GraphQL schema across multiple OpenApi3 specs?


JHibbard:
Can oasgraph generate one GraphQL schema across multiple OpenAPI3 specs?  I have several OpenAPI3 compliant services with links that reference each other.  It'd be amazing to use GraphQL to query across them.

ErikWittern:
@JHibbard This feature is not yet supported by OASGraph, but falls neatly into its intended capabilities. One challenge will be to avoid collision between GraphQL type names. Another consideration is whether OASGraph should be passed a list of OAS, or whether a single OAS that defines links using `operationRef`s could be the starting point.

amnuts:
I would also be really interested in seeing this functionality!  I have several microservices which each have their own OpenAPI spec and would like to put a GraphQL gateway in front of them.  Being able to supply all the oas specs and have it produce one unified GraphQL schema would be fantastic.

ErikWittern:
@amnuts We are currently fixing some issues for a new minor version (which will come out soon), and then want to look into this one!

We will likely approach this in different stages, i.e., initially throw errors if there is collusion between the type names used in the OAS, and eventually mitigate these problems.

amnuts:
That's great to hear @ErikWittern, thanks!  Right now we're kind of working around by converting each of the OAS files in turn with oasgraph and then using graphql-tools to merge the schemas together in the end.  But I feel that's probably going to limit us by not allowing the 'links:' from one spec to another - not sure, yet.

But thank you so much for this project!  It was amazing to be able to use an OAS file and have it converted automatically to GraphQL including all of the resolvers and mutations, which meant we only needed to concentrate on building a good foundation spec.  Brilliant!

Alan-Cha:
@JHibbard @amnuts Yesterday I added support for multiple OASs! I am still working on documentation and other improvements but if you would like to give it try, please do! 

It would also be nice just to have some early feedback so I know what to do next because as of now, we only have a basic version with limited collision handling (if `operationIds` or `securitySchemes` are the same, OASGraph will simply pick from one OAS and provide an error message. To mitigate this, you may need to use the new [`operationIdFieldNames`](https://github.com/strongloop/oasgraph/tree/master/packages/oasgraph#options) option).

***

Now, instead of providing a Swagger or an OAS, you can _also_ provide an array of a mix of Swaggers and OASs like in the following: 

```
OasGraph.createGraphQlSchema([oas1, oas2, oas3])
```

[...]

PR #129, opened 2019-03-14
Links between multiple OAS

Alan-Cha:
Related to #82
Supports OAS title reference
Need to add other support

Handled #128
By refactoring generate operationId func


packages/oasgraph/src/oas_3_tools.ts
```diff
@@ -1036,4 +1036,11 @@ export function capitalize(str: string): string {
  */
 export function uncapitalize(str: string): string {
   return str.charAt(0).toLowerCase() + str.slice(1)
+}
+
+/**
+ * For operations that do not have an operationId, generate one
+ */
+export function generateOperationId(method: string, path: string): string {
+  return ((beautify(`${method}:${path}`) as any) as string)
```
ErikWittern: Strange type conversion. Is that really necessary?

Alan-Cha: Hmmm, you're right... I refactored old code and I didn't think much of it afterwards. I'll change it.

ErikWittern:
@Alan-Cha Great work, getting closer to GraphQL-based API composition!

PR #130, opened 2019-03-15
Added documentation

Alan-Cha:
I think we should consider putting in more documentation about links in general in the README.md.


Alan-Cha:
I tried to keep it as short as possible but also give the reader a good idea of the capabilities.

ErikWittern:
@Alan-Cha Perfect!

PR #132, opened 2019-03-18
Adding multi OAS to CLI

Alan-Cha:


packages/oasgraph-cli/src/oasgraph.ts
```diff
@@ -30,41 +27,64 @@ program
...
+      getRemoteFileSpec(filePath)
+      .then((remoteContent) => {
+        resolve(remoteContent)
+      })
+      .catch((error) => {
+        console.error(error)
+        reject(filePath)
+      })
+
+      return
```
ErikWittern: You don't have to return here or in line 58 given the use of the `if ... else` block.

Alan-Cha: Ah right, good call.

ErikWittern:
@Alan-Cha Great, thanks for the changes.

PR #133, opened 2019-03-26
Consistency

### LINKED ISSUE #119: FAIL  test/example_api.test.js on master
Hello guys,

The test/example_api.test.js is currently broken on master. Stack (npm run test):

`● Link parameters as constants and variables with request payload

    expect(received).toEqual(expected)

    Difference:

    - Expected
    + Received

      Object {
    -   "data": Object {
    -     "postScanner": Object {
    -       "body": "req.body: body, req.query.query: query, req.path.path: path",
    -       "everythingLink2": Object {
    -         "body": "http://localhost:3002/api/scanner/path_post_200_body_query_path_application/json_req.body: body, req.query.query: query, req.path.path: path_query_path_close",
    -       },
    -     },
    -   },
    +   "errors": Array [
    +     [GraphQLError: Cannot query field "everythingLink2" on type "Scanner". Did you mean "everythingLink"?],
    +   ],
      }

      286 |   }`
      287 |   return graphql(createdSchema, query).then(result => {
    > 288 |     expect(result).toEqual({
          |                    ^
      289 |       data: {
      290 |         "postScanner": {
      291 |           "body": "req.body: body, req.query.query: query, req.path.path: path",

      at toEqual (test/example_api.test.js:288:20)
`

Alan-Cha:
Removed addSubOperations
Collapsed links
Links are now stored in a `dataDef` rather than as a part of the preprocessing data
We do not sort by number of links (and subOperations) anymore, rather, just on array return type 

Fixed #119 
Now we run Travis tests for Node 11 as well 
On a side note, why are we running Node 8 and Node 10 but not 9?

Todo: Create object types first, then reuse them with different resolvers


packages/oasgraph/lib/index.d.ts
```diff
@@ -11,7 +11,7 @@
  * - GraphQL (Input) Object Types must have a unique name. Thus, sometimes Input
  *   Object Types and Object Types need separate names, despite them having the
  *   same structure. We thus append 'Input' to every Input Object Type's name
- *   as a convention.
+ *   as a convention.67890
```
ErikWittern: Not sure about this one... how about 67889?

packages/oasgraph/src/index.ts
```diff
@@ -16,7 +16,7 @@
  * - GraphQL (Input) Object Types must have a unique name. Thus, sometimes Input
  *   Object Types and Object Types need separate names, despite them having the
  *   same structure. We thus append 'Input' to every Input Object Type's name
- *   as a convention.
+ *   as a convention.67890
```
ErikWittern: See above.

packages/oasgraph/src/index.ts
```diff
@@ -166,6 +161,15 @@ async function translateOpenApiToGraphQL (
...
 
+//   const fs = require('fs');
+// fs.writeFile("preprocessing.json", JSON.stringify(data, null, 2), function(err) {
+//     if(err) {
+//         return console.log(err);
+//     }
+
+//     console.log("The file was saved!");
+// }); 
+
```
ErikWittern: Maybe remove this.

.travis.yml
```diff
@@ -3,6 +3,7 @@ language: node_js
 node_js:
   - '8'
```
ErikWittern: Let's test version 9 also, as you suggested.

Alan-Cha:
The last commit creates the object types first and then adds resolvers later.

Alan-Cha:
Actually, I think there is still more work to be done. Ideally, I should create the object types by cycling through the `dataDef`s. I think the last commit is still a good intermediary commit.

Alan-Cha:
I think the best case scenario is we cycle through the `dataDef`s and create the object types with the need of creating new `dataDef`s or reusing preexisting object types. However, I do not think this is possible without making significant changes to the preprocessing step, naming in addition to what we already do, breaking down schemas into their component parts. We can continue with the pattern of simultaneous creating and reusing for now but we should try to cycle through the `dataDef`s instead of the operations because this is closer to how we understand how OASGraph should work.

Alan-Cha:
Closed because it is redundant. Check #138.

Issue #117, opened 2019-03-01
Add "requestOptions" to createGraphQLSchema's options


thejibz:
Hi guys it's me again :)

I encountered the need to set a proxy to query a backend. The simplest workaround was:
```
//` make the call
    log(`Call ${options.method.toUpperCase()} ${options.url}` +
      `?${querystring.stringify(options.qs)} ` +
      `headers:${JSON.stringify(options.headers)}`)
     
      options.proxy = "myproxy:3128"

    return new Promise((resolve, reject) => {
         request(options, (err, response, body) => {
        if (err) {
```
So I think it could be useful to allow user to specify extra options for the nodejs' request object. Furthermore it is often the case for other lib.

Something along the line of:
`Object.assign(options, requestOptions)`

What's your take on that ?

Alan-Cha:
@thejibz This is a great idea! One concern is that there may be some overlap with preexisting options like `headers` and `qs`. Should we assign some kind of priority or get rid of them entirely in favor of `requestOptions`? Thoughts @ErikWittern?

thejibz:
Regarding overlaping, I agree that these options kind of conflict... Maybe as a first step just add a simple "proxy" option if other request's configs seems non-relevant ? I just made a fork with the new "requestOptions" parameter, it works, but I will wait on a consensus on what to do before making a PR.

Alan-Cha:
I think going for the full `requestOptions` is a good idea. I would like @ErikWittern to weigh in but feel free to submit a PR for us to discuss and iterate on! 😄

ErikWittern:
I agree, this is a good idea. A PR would be highly appreciated.

Alan-Cha:
Accepted PR #135

PR #135, opened 2019-04-02
Add requestOptions for createGraphQlSchema method

thejibz:
Hello guys here is the PR ! Related to #117 

I also added a test for it in "autentication.test.js" where I use the new "requestOptions" instead of the "header" one to set the credentials.

Signed-off-by: JB Rx <jb.roux@ymail.com>


packages/oasgraph/src/resolver_builder.ts
```diff
@@ -181,6 +185,44 @@ export function getResolver ({
       qs: query
     }
 
+    /**
+     * Deep merge optional "requestOptions" parameter with current request options.
+     * See https://github.com/TehShrike/deepmerge
+     */
+    if (!!requestOptions) {
```
Alan-Cha: Is this double negation necessary?

packages/oasgraph/src/resolver_builder.ts
```diff
@@ -331,7 +373,7 @@ export function getResolver ({
  * Attempts to create an object to become an OAuth query string by extracting an
  * OAuth token from the ctx based on the JSON path provided in the options.
  */
-function createOAuthQS (
+function createOAuthQS(
```
Alan-Cha: I believe in our style convention, we have a space between the function name and the parameters. Would you mind reverting these changes? There are a few more in the PR I believe.

packages/oasgraph/src/resolver_builder.ts
```diff
@@ -510,7 +552,7 @@ function resolveLinkParameter(paramName: string, value: string, resolveData: any
     if (value === '$request.body') {
       return resolveData.usedPayload
 
-    // CASE: parameter in previous body
+      // CASE: parameter in previous body
```
Alan-Cha: I believe in our style convention, we usually keep comments indented the same as the surrounding code. Would it be possible to move these back?

packages/oasgraph/src/types/options.ts
```diff
@@ -89,6 +92,13 @@ export type Options = {
    * operationId. 
    */
   operationIdFieldNames?: boolean
+
+  /**
+   * Allow to override or add options to the node's request object used to make
```
Alan-Cha: I am in favor of changing "Allow" into "Allows" to match the tense of the other comments. What do you think?

packages/oasgraph/src/types/options.ts
```diff
@@ -165,4 +175,11 @@ export type InternalOptions = {
    * operationId. 
    */
   operationIdFieldNames: boolean
+
+  /**
+   * Allow to override or add options to the node's request object used to make
```
Alan-Cha: Same as above.

packages/oasgraph/src/resolver_builder.ts
thejibz: It's considered a best praticte by some (see http://airbnb.io/javascript/#coercion--booleans) to express that we only care about the boolean outcome.

But yes, in the case of an if statement it's not necessary ! It's just a habit of mine to use it :D I will change it.

Alan-Cha: Oh interesting! I learned something new today. I will look into this in my own time. However, I believe we should remove it as this is the first time this pattern has been used in our code base. Hope you understand!

packages/oasgraph/src/types/options.ts
thejibz: +1

packages/oasgraph/src/types/options.ts
thejibz: +1

packages/oasgraph/src/resolver_builder.ts
thejibz: I will change the vscode linter rules for this project. Hope it is possible, if not I will install jslint.

packages/oasgraph/src/resolver_builder.ts
thejibz: I will change my vscode linter rules for this project. Hope it is possible, if not I will install jslint.

packages/oasgraph/src/resolver_builder.ts
```diff
@@ -181,6 +185,44 @@ export function getResolver ({
       qs: query
     }
 
+    /**
```
Alan-Cha: I am concerned about `deepmerge`. I wonder if it is necessary. `deepmerge` is best suited for complex objects and because options, especially at this point in the code, only contains strings or hashes of strings, I think we can do the same amount of work by just using `Object.assign()`. 

As a side note, I am also not sure if I agree that `requestOptions` should have priority. 

Something along the lines of the following (I wrote it quickly so forgive me if there are mistakes)

```
let options: RequestOptions 
if (requestOptions) {
  options = requestOptions

  options['method'] = operation.method
  options['url'] = url
  if (options.headers) {
    Object.assign(options.headers, headers)
  } else {
    options['headers'] = headers
  }
  if (options.qs) {
    Object.assign(options.qs, query)
  } else {
    options['qs'] = query
  }
} else {
  options = {
    method: operation.method,
    url: url,
    headers: headers,
    qs: query
  }
}
```

We would like to minimize our dependencies so if we can avoid using them without writing too much code, then I think we should. Let me know your thoughts.

Alan-Cha:
@thejibz Thank you so much for the pull request. Sorry, I should have added all of my comments at once before letting you see them. 

As you have already know, there are a few concerns with the style. I am also concerned about the `deepmerge`. Aside from that, everything looks really good and nicely commented! This is great work!

***

Just one other thing, we need to add documentation in the [README.md](https://github.com/strongloop/oasgraph/tree/master/packages/oasgraph#options)

Here is my suggestion. Feel free to improve it. 

`requestOptions` (type: `object`, default: `{}`): Additional [options](https://github.com/request/request#requestoptions-callback) that can be used to configure the HTTP calls, from the [`Request module`](https://github.com/request/request), that powers the GraphQL resolvers. 

Raw (copy/paste) text:
```
`requestOptions` (type: `object`, default: `{}`): Additional [options](https://github.com/request/request#requestoptions-callback) that will be used to configure the HTTP calls, from the [`Request module`](https://github.com/request/request), that powers the GraphQL resolvers. 
```

***

We will also need to remove the overlapping options at a later date but I think this is a step in the right direction. Thank you so much once again @thejibz. This is all greatly appreciated!

packages/oasgraph/src/resolver_builder.ts
thejibz: Ha ! In my original work I used Object.assign(options, requestOptions) ! But doing so you would overwrite headers so I thought that for headers a "combine arrays" approach will be better than overwriting. Hence the deepmerge.

But with your approach you avoid this by explicitly assigning the headers, so I'm fine with it :)

Regarding your snippet, I'm fine with the approach ie. giving priority to options already set.

So if there is no other remarks I will amend the PR with a code similar to your snippet.

packages/oasgraph/src/resolver_builder.ts
thejibz: Ok it's done and will be shipped in the next commit.

packages/oasgraph/src/resolver_builder.ts
thejibz: Ok it's done and will be shipped in the next commit.

packages/oasgraph/src/resolver_builder.ts
Alan-Cha: I would like @ErikWittern to weight in but other than these changes, I am happy with what we have.

thejibz: Ok, let's wait some more discussion then :)

ErikWittern: I like the agreed-upon approach, where priority of fields is considered per case. I think line 3 should be `options = {...requestOptions}` (or use `Object.assign`) to not permanently overwrite `requestOptions` in the following lines?!

Alan-Cha: Hmmm I think the same fields will be overwritten in all cases so it doesn't matter if we preserve `requestOptions`. I suppose it doesn't hurt just to do it...

```
let options: RequestOptions
if (requestOptions) {
  options = {...requestOptions}

  options['method'] = operation.method
  options['url'] = url
  if (options.headers) {
    Object.assign(options.headers, headers)
  } else {
    options['headers'] = headers
  }
  if (options.qs) {
    Object.assign(options.qs, query)
  } else {
    options['qs'] = query
  }
} else {
  options = {
    method: operation.method,
    url: url,
    headers: headers,
    qs: query
  }
}

```

Alan-Cha: @thejibz If you can do this change and add the documentation like I mentioned in the PR comment, I think we'll be ready to accept!

thejibz: @Alan-Cha , on my way for the PR :) Hope to deliver today.

thejibz: > I like the agreed-upon approach, where priority of fields is considered per case. I think line 3 should be `options = {...requestOptions}` (or use `Object.assign`) to not permanently overwrite `requestOptions` in the following lines?!

Sounds good, doesn't work. Sorry for this "trait d'humour", in fact the Node Request's Options isn't fully compatible with the OASGraph "RequestOptions". E.g, in Node Request: "method" is optional, url is of type Url and not String...

So I think about amend the PR in the way that "RequestOptions" is no longer defined by OASGraph but taken directly from Node Request's Options type. Let me know if it works out for you ! The other way around is modifying "RequestOptions" to fully respect Node Request's options

thejibz: The could would be:
```
import * as NodeRequest from 'request'
// ...
let options: NodeRequest.OptionsWithUrl
    if (requestOptions) {
      options = { ...requestOptions }

      options['method'] = operation.method
      options['url'] = url
      if (options.headers) {
        Object.assign(options.headers, headers)
      } else {
        options['headers'] = headers
      }
      if (options.qs) {
        Object.assign(options.qs, query)
      } else {
        options['qs'] = query
      }
    } else {
      options = {
        method: operation.method,
        url: url,
        headers: headers,
        qs: query
      }
    }
```

ErikWittern:
@thejibz Really good work, thank you so much for this! One issue, though, is that `deepmerge` is still requested as a dependency, despite it not being used, it seems?!

If that is the case, could you please remove that dependency before we merge the PR?

packages/oasgraph/src/resolver_builder.ts
Alan-Cha: This is a great change! Thank you for doing the investigation work!

thejibz:
> 
> 
> @thejibz Really good work, thank you so much for this! One issue, though, is that `deepmerge` is still requested as a dependency, despite it not being used, it seems?!
> 
> If that is the case, could you please remove that dependency before we merge the PR?

You're welcome ! Ooops for "deepmerge", yes I amend the PR tomorrow. Time for some beers now ;)

ErikWittern:
@thejibz Fantastic, thank you! I hope the beers were delicious!

PR #142, opened 2019-05-16
add cookie authentication to builder

jduffy3:


Alan-Cha:
Great work! I see you have made other improvements in the code base like removing unused imports and making code clearer and the formatting more consistent. This is greatly appreciated!

Just a few concerns, however...

1. Would you add a test case?
The test case should be added [here](https://github.com/strongloop/oasgraph/blob/master/packages/oasgraph/test/authentication.test.js). I believe we will also need to make some changes [here](https://github.com/strongloop/oasgraph/blob/master/packages/oasgraph/test/example_api_server.js#L210) so that the example API can utilize the new cookie-base API key.

2. Would you squash your commits? 

```
git rebase -i HEAD~3
```

3. Please make sure all of your commits are properly signed off.

Thanks! 😄

jduffy3:
Cool added tests, squashed commits and commits should be signed!

Alan-Cha:
Thank you so much, @jduffy3! Great work! 😄

Issue #140, opened 2019-05-14
fillEmptyResponses


michaelsearles:
**Describe the bug**
We have POST and Delete operations that do not return any JSON responses and hence we do not define any response schema in the v2 OAS file in these particular cases. This causes warning in the console when we try to generate the wrapper using OASGraph.

***********************************************************************************************
"warnings": [
    {
      "type": "MissingResponseSchema",
      "message": "Operation 'POST /v1/persons/{concern_role_id}/special_cautions' has no (valid) response schema. You can create placeholder schemas using the fillEmptyResponses option.",
      "mitigation": "Ignore operation"
    },
    {
      "type": "MissingResponseSchema",
      "message": "Operation 'DELETE /v1/persons/{concern_role_id}/special_cautions/{special_caution_id}' has no (valid) response schema. You can create placeholder schemas using the fillEmptyResponses option.",
      "mitigation": "Ignore operation"
    },
    {
      "type": "MissingResponseSchema",
      "message": "Operation 'POST /v1/outcome_plans/{outcome_plan_id}/attachments' has no (valid) response schema. You can create placeholder schemas using the fillEmptyResponses option.",
      "mitigation": "Ignore operation"
    },
  
............

***********************************************************************************************

I have tried setting the option "fillEmptyResponses: true" and this does substantially reduce the number of warnings especially Delete operations, but some still remain.

***********************************************************************************************
"warnings": [
    {
      "type": "MissingResponseSchema",
      "message": "Operation 'POST /v1/persons/{concern_role_id}/special_cautions' has no (valid) response schema. You can create placeholder schemas using the fillEmptyResponses option.",
      "mitigation": "Ignore operation"
    },
    {
      "type": "MissingResponseSchema",
      "message": "Operation 'POST /v1/outcome_plans/{outcome_plan_id}/attachments' has no (valid) response schema. You can create placeholder schemas using the fillEmptyResponses option.",
      "mitigation": "Ignore operation"
    },

........

***********************************************************************************************

**Expected behavior**
Setting the option to fill empty responses should fill all REST resources with empty responses 

**Screenshots**

**Additional context**

Alan-Cha:
Hmmm... this is unexpected behavior. Is it possible to share the OAS (Swagger 2?) or a portion of it? If not, is it possible for you to compare the operations that were affected by the `fillEmptyResponses` option with those that weren't and let us know?

michaelsearles:
[...]

PR #143, opened 2019-05-24
Small change to warning message

Alan-Cha:
Related #140


ErikWittern:
@Alan-Cha Great, better docs are always welcome!

PR #147, opened 2019-06-05
Prettier and const

Alan-Cha:


ErikWittern:
@Alan-Cha Very pretty!

PR #148, opened 2019-06-06
General fixes

Alan-Cha:


packages/oasgraph/test/government_social_work_api.test.js
```diff
@@ -75,10 +75,3 @@ test('Get resource', () => {
   let errors = validate(createdSchema, ast)
   expect(errors).toEqual([])
 })
-
-test('Strict mode throws exception', () => {
```
Alan-Cha: This test does not do anything... LinkNameCollision error was not used so an error was never thrown. I think we will need to add a new test for strict mode.

Alan-Cha:
Removed CHANGES.md because I don't think we're ever going to use it. It's just stagnating. 

In a recent commit, I changed all concatenated error messages to use double quotes for consistency (we were using single and double throughout). Unfortunately, double quotes doesn't always print nicely so I changed them all to single quotes.

packages/oasgraph/src/index.ts
```diff
@@ -447,3 +449,75 @@ function sortOperations(op1: Operation, op2: Operation): number {
...
+      return titles.indexOf(title) !== index
+    })
+  ).forEach(title => {
+    handleWarning({
+      typeKey: 'MULTIPLE_OAS_SAME_TITLE',
+      culprit: title,
+      data,
+      log: translationLog
+    })
+  })
```
ErikWittern: Could this not be much easier? I.e., use `const titleSet = new Set(oass.map(oas => oas.info.title)` and then `if (oass.length > titleSet.size) { // duplicate!`?

packages/oasgraph/src/schema_builder.ts
```diff
@@ -472,100 +472,111 @@ function createFields({
...
-        // finally, add the object type to the fields (using sanitized field name)
-        const saneLinkKey = Oas3Tools.beautifyAndStore(linkKey, data.saneMap)
-        // TODO: check if fields already has this field name
-        fields[saneLinkKey] = {
-          type: resObjectType,
-          resolve: linkResolver,
-          args,
-          description
+          // finally, add the object type to the fields (using sanitized field name)
+          Oas3Tools.beautifyAndStore(saneLinkKey, data.saneMap)
```
Alan-Cha: We don't need to beautify anymore because the links should already have sanitized keys. I kept it here for now because I'm not sure how the store will affect other parts of the code.

packages/oasgraph/src/index.ts
ErikWittern: Well, you would not have the `culprit` then...

packages/oasgraph/src/preprocessor.ts
```diff
@@ -382,17 +382,24 @@ export function createDataDef(
...
   if ('allOf' in schema) {
     schema = mergeAllOf(schema)
   }
 
+  const preferredName = getPreferredName(names)
+
+  const saneLinks = {}
+  if (typeof links === 'object') {
+    Object.keys(links).forEach(linkKey => {
+      saneLinks[Oas3Tools.beautify(linkKey)] = links[linkKey]
```
Alan-Cha: We sanitize link keys now because I discovered a `productLead` and a `ProductLead` link definition in the Example API OAS. We would not report this and could have unintended/unreported consequences.

packages/oasgraph/test/instagram.test.js
```diff
@@ -38,10 +38,3 @@ test('All Instagram query endpoints present', () => {
   ).length
   expect(gqlTypes).toEqual(oasGetCount)
 })
-
-test('Strict mode throws exception', () => {
```
Alan-Cha: Same thing goes for here.

packages/oasgraph/test/government_social_work_api.test.js
ErikWittern: All right.

packages/oasgraph/src/preprocessor.ts
ErikWittern: Good find.

PR #154, opened 2019-06-08
Add addSlicingPattern option

Alan-Cha:
Fix bug with nested object lists


packages/openapi-to-graphql/README.md
```diff
@@ -164,6 +164,8 @@ The options object can contain the following properties:
...
 
+- `addSlicingPattern` (type: `boolean`, default: `false`): [Pagination](https://graphql.org/learn/pagination/) allows for control over fields that return lists of objects. There are different pagination models such as the slicing pattern, which adds arguments to fields, and the connections pattern, which in addition to new arguments also introduces a number of new types and offers finer control. This option will follow the former model, adding `first` and `last` arguments to fields that return lists of objects.
```
ErikWittern: I am not sure whether we even want to mention the connections pattern here.

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -418,6 +420,50 @@ export function getResolver({
             }
           }
 
+          // Apply autogenerated slicing arguments
+          if (data.options.addSlicingPattern && Array.isArray(saneData)) {
+            // NOTE: Does not differentiate between autogenerated args and
+            // preexisting args
+            if ('first' in args && 'last' in args) {
```
ErikWittern: So, we are relying on a convention here, namely that the slicing arguments are named `first` or `last`. I think we can do more research on common names in our GitHub dataset. But even then, maybe the user needs to provide names of slicing arguments?

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -418,6 +420,50 @@ export function getResolver({
...
+                new Error(
+                  `Cannot use both 'first' and 'last' slicing arguments at the same time`
+                )
+              )
+            } else {
+              let arraySaneData = saneData as Array<any>
+
+              if ('first' in args) {
+                if (args['first'] >= 0) {
+                  arraySaneData = arraySaneData.slice(0, args['first'])
```
ErikWittern: Hm, I am wondering about this: By cutting the array returned from the REST API, I think we easily confuse users, because it is now unclear whether, for example, `last` relates to the last n entries of the first q entries fetched from the REST API, or to the last entries of all the data?

Alan-Cha:
@ErikWittern I adapted the following [specification](https://facebook.github.io/relay/graphql/connections.htm#sec-Pagination-algorithm). Does this change your opinion?

packages/openapi-to-graphql/README.md
Alan-Cha: Connections pattern is something I want to look into so I thought I might as well mention it here now. 

On the topic of connections pattern, I feel like I should mention implementing the connections pattern will be quite a daunting task. First of all, it will probably have a huge affect on how we handle resolvers and link resolvers. Secondly, it raises a bunch of open questions: how do we implement a `cursor`? I think we need knowledge about the object’s `id`. We can pass the responsibility to the user but they would need to take care of it through the `customResolvers` option but the way it is currently set up (i.e. using the path and the method) does not make it easy to manipulate the resolvers kf the `edges` and `nodes` types.

packages/openapi-to-graphql/src/resolver_builder.ts
Alan-Cha: As mentioned, I provided a link to a connections pattern specification provided by Facebook. 

If we deem it okay to adapt the specification to suit the slicing pattern, then what I have implemented should be fine because according to the specification, `first` and `last` affects the returned data, not the ends of the actual data. 

Something else to keep in mind is I’m hoping to use the slicing pattern as a stepping stone to implement the connections pattern.

packages/openapi-to-graphql/README.md
ErikWittern: I fully agree. Implementing the connections pattern will in many cases be impossible, IMO. Because it depends on certain features (slicing arguments with `first` and `last` semantic, `cursors` etc.) to be exposed by the underlying API. I am not at all confident we will be able to support this for arbitrary APIs in OASGraph.

packages/openapi-to-graphql/src/resolver_builder.ts
ErikWittern: I am still not on board with the semantics of `first` and `last` you introduce here. They relate to (arbitrary) slices of data returned by the underlying REST call. In contrast, I think a user of a GraphQL native API would expect these to return the `first` or `last` entries of all data, or of a slice of data defined by a `cursor`.

ErikWittern: I.e., I would be much more in favor of implementing slicing using a `limit` argument, which does not imply _which_ part of a list of data is returned.

Alan-Cha:
See #155

PR #155, opened 2019-06-11
Limit argument

Alan-Cha:


packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -111,14 +111,6 @@ export function getResolver({
       resolveData.usedParams = {}
     }
 
-    /**
```
Alan-Cha: I think this is leftover code from `subOperations`.

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -39,20 +40,19 @@ type AuthOptions = {
...
 }
+
 /**
  * Creates and returns a resolver function that performs API requests for the
  * given GraphQL query
  */
 export function getResolver({
   operation,
   argsFromLink = {},
-  argsFromParent = [],
```
Alan-Cha: `subOperations`. See below.

packages/openapi-to-graphql/README.md
```diff
@@ -164,6 +164,8 @@ The options object can contain the following properties:
...
 
+- `addLimitArgument` (type: `boolean`, default: `false`): Add a `limit` argument to fields that return lists of objects/lists that will limit the number of returned elements.
```
ErikWittern: Maybe add to the end: `..., selecting the first n elements of the list.` ? I.e., I think it is important to clearly lay out the semantics of this argument.

packages/openapi-to-graphql/lib/types/options.d.ts
```diff
@@ -112,6 +112,15 @@ export declare type Options = {
             };
         };
     };
+    /**
+     * Auto-generate a 'limit' argument for all fields that return lists of
+     * objects, including ones produced by links
+     *
+     * Allows to manage the return size of lists of objects
```
ErikWittern: `manage` -> `constrain`?

packages/openapi-to-graphql/README.md
ErikWittern: You use `x` (not `n`) below, we should be consistent - so maybe use `x` here as well?! (Even though I slightly favor `n`, because it is often associated with a positive integer value.)

packages/openapi-to-graphql/src/resolver_builder.ts
ErikWittern: Good to go, then.

packages/openapi-to-graphql/src/schema_builder.ts
```diff
@@ -889,6 +905,33 @@ export function getArgs({
...
+      handleWarning({
+        typeKey: 'LIMIT_ARGUMENT_NAME_COLLISION',
+        culprit: `${operation.method.toLocaleUpperCase()} ${operation.path}`,
+        data,
+        log: translationLog
+      })
+    } else {
+      args['limit'] = {
+        type: GraphQLInt,
+        description: `Auto-generated argument that limits size of returned list of objects/lists`
```
ErikWittern: Here, I think we need to convey the semantics of limit (`Selects the first n elements from a list`) to the end-user!

packages/openapi-to-graphql-cli/README.md
```diff
@@ -19,14 +19,15 @@ npm i -g openapi-to-graphql-cli
 ## Usage
 
 ```
-Usage: openapi-to-graphql <OAS JSON file path(s) and/or remote url(s)> [options]
+Usage: openapi-to-graphql <OAS JSON file path(s) and/or remotlee url(s)> [options]
```
Alan-Cha: Oops will fix this tomorrow.

ErikWittern:
@Alan-Cha This opens limitless possibilities!

PR #158, opened 2019-06-12
Sanitize

Alan-Cha:


packages/openapi-to-graphql/src/index.ts
```diff
@@ -539,3 +539,5 @@ function preliminaryChecks(
     })
   }
 }
+
+export { sanitize } from './oas_3_tools'
```
Alan-Cha: Users probably want this, right?

packages/openapi-to-graphql/src/oas_3_tools.ts
```diff
@@ -466,7 +468,7 @@ export function inferResourceNameFromPath(path: string): string {
     }
   })
 
-  return sanitize(pathNoPathParams)
```
Alan-Cha: We run `sanitize` in `createDataDef`. Before then, all potential names are not sanitized, so sanitizing here (early) seems to be redundant.

packages/openapi-to-graphql/src/oas_3_tools.ts
```diff
@@ -493,19 +495,19 @@ export function getRequestSchema(
...
-      // Prioritizes content-type JSON
+      // Prioritize content-type JSON
       if (Object.keys(content).includes('application/json')) {
         return {
           payloadContentType: 'application/json',
           payloadSchema: content['application/json'].schema as SchemaObject
         }
       } else {
-        // Picks a random content type
-        for (let contentType in content) {
```
Alan-Cha: I think this code might be confusing/bad

packages/openapi-to-graphql/src/oas_3_tools.ts
```diff
@@ -624,19 +625,19 @@ export function getResponseSchema(
...
-        // Prioritizes content-type JSON
+        // Prioritize content-type JSON
         if (Object.keys(content).includes('application/json')) {
           return {
             responseContentType: 'application/json',
             responseSchema: content['application/json'].schema as SchemaObject
           }
         } else {
-          // Picks a random content type
-          for (let contentType in content) {
```
Alan-Cha: Similar change as above.

ErikWittern:
@Alan-Cha Beautiful and sane!

PR #159, opened 2019-06-13
Add createOperationString

Alan-Cha:


Alan-Cha:
Moved `oass` into `data`
Created `createOperationString` and `formatOperationString` functions

All in all, simplified function signatures and made messages (referencing a specific operation) more consistent

ErikWittern:
@Alan-Cha Thank you! The code-base is improving bit by bit.

PR #172, opened 2019-06-20
Add deprecation message

Alan-Cha:


ErikWittern:
@Alan-Cha Very good! We may need to publish a patch to npm also, to reflect this change.

PR #178, opened 2019-06-26
Support content property in parameter objects

### LINKED ISSUE #153: Support OAS content property
According to 3.0.2 and in the context of [parameter objects](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#parameterObject), "For more complex scenarios, the content property can define the media type and schema of the parameter. A parameter MUST contain either a schema property, or a content property, but not both."

Currently, we only support the schema property and not the content property. I'm not sure if other parts of OAS has followed suit.

Alan-Cha:
Fix #153


ErikWittern:
@Alan-Cha Great work, as always!

PR #184, opened 2019-06-28
Fix #183 field creation should use prop name instead of ref name

### LINKED ISSUE #183: Property with ref in "definitions" resolved to use ref name instead of prop name
Actually I am not sure whether this supports v2 or v3 spec?

**Describe the bug**
OAS (:
```json
{
  ...
  "paths": {
    "/my_api": {
      "get": {
        "responses": {
          "200": {
            "schema": {  "$ref": "#/definitions/Response" }
          },
          "products": [ "application/json" ]
        }
      }
    },
    ...
  },
  "definitions": {
    "Response": {
      "type": "object",
      "properties": {
        "data": {
          "type": "object",
          "$ref": "#/definitions/Item"
        }
      }
    },
    "Item": {
      "type": "object",
      "properties": {
        "id": {
          "type": "number"
        }
      }
    }
  }
}
```

Will be resolved to graphql:
```graphql
type Query {
  getMyQuery: Response
}
type Response {
  item: Item
}
type Item {
  id: Int
}
```

**To Reproduce**
1. createGraphQlSchema with above schema
2. check in GraphiQL document
or use cli to generate graphql file

**Expected behavior**
```graphql
type Query {
  getMyQuery: Response
}
type Response {
  data: Item
}
type Item {
  id: Int
}
```

**Screenshots**
-

**Additional context**
-

JennieJi:
Actually shall we ignore the built files?


ErikWittern:
@JennieJi Thank you for this fix! We will publish a patch version incorporating your change.

We currently keep the build files in the repository to be able to publish more easily to npm.

JennieJi:
Glad to hear this.

For built files, I feel safer to add script `prepublish` or `prepublishOnly` with `npm run build`, so when running `npm publish` will always build the latest version.

ErikWittern:
@JennieJi Good point about the `prepublish` / `prepublishOnly` scripts. However, we use `lerna` to publish new versions, not `npm` itself. Do you know if a similar setup of scripts is also possible for lerna?

Issue #181, opened 2019-06-27
Singular names from plural URLs with IDs


Alan-Cha:
A common trope in API design is to offer an operation that will return an array of a certain type of object, and another that will return a singular instance if given an ID. 

For example: `GET exampleAPI.com/apis/v1/users` and `GET exampleAPI.com/apis/v1/users/{id}`

It would be a nice addition if we could make these names singular. 

For example: `GET exampleAPI.com/apis/v1/users` will create a `users` field and `GET exampleAPI.com/apis/v1/users/{id}` will create a `user` field.

***

There's a few things to consider here. 
* It seems to be a common to add the `id` parameter immediately after the path element in question. 
* IDs come in different shapes and forms. Some are just `id`, some have a single letter appended to them like `pid`, and some of them have the singular name appended to them like `userId`
* To get the singular name, we might need to use some kind of dictionary or stemming library. 
  * We might need to tokenize the path element so that the library can read it (consider different cases like camelCase or python_case)

Alan-Cha:
After pulling in #296, we now have support for this kind of feature through the`singularNames` option.

Issue #180, opened 2019-06-27
Extend basePath


Alan-Cha:
The `basePath` is a part of the URL. It sits between the server target host URL and the operation path. An improvement we can do, in regards to creating better names from paths (see #179), is to determine the lowest common denominator `basePath` for all operation paths. That way, we can build names from only the unique components of the path and not mistakenly add redundant metadata.

For example, a common trope is to add the API version number to the path. 
```
exampleAPI.com/apis/v1/...
```
We can filter out the `basePath` `apis/v1/` and strictly focus on the semantically important parts of the path.

Alan-Cha:
The `basePath` seems to be an element in Swagger but not so in OAS 3.x.x. Regardless, we can still use the same strategy.

Issue #179, opened 2019-06-26
Improving Query and Mutation field names


Alan-Cha:
Determining good names in OpenAPI-to-GraphQL is a perpetual problem because [OAS](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md) is not designed to provide the information we need. Therefore, the best we can do is to assume the API well-designed and the OAS is written in an ideal/idiomatic way and provide numerous mechanisms to produce names if any of them fail. 

It is impossible to create a solution that will create good names for all APIs and all OASs. That is why we have the [`operationIdFieldNames`](https://github.com/IBM/openapi-to-graphql/tree/master/packages/openapi-to-graphql#options) option, which allows the user to easily define custom field names manually. This issue is focused on what the default behavior should be given the best conditions. 

***

The current methodology is as follows. Query fields, generally speaking, borrow their names from their respective object types. For example, the field `Query.user` may return a `User` object. 

We think this strategy allows us to separate from similar projects. We are more data-centric while alternative solutions are more operation-centric. We believe that being data-centric allows us to align more closely with the intent of GraphQL. We are able to reuse object types and create complex nested object type structures. 

However, a problem occurs when multiple operations return the same field. Because field names must be unique, only one operation will adopt the name of the data type. The others will default to their `operationId`. As a result, the GraphQL interface will be non-deterministic and inconsistent. 

For example, let's say there are two operations `GET /coffeeShopLocation` and `GET /sandwichShopLocation` that return the same kind of location-related data type. The schema of the data type is stored in the [`components`](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#components-object) field so it is referenced by both operations and OpenAPI-to-GraphQL knows to reuse the generated data type. In an ideal situation, we will produce a `coffeeShopLocation` and `sandwichShopLocation` fields, which both return a `Location` object type.

Currently, this does not happen. Instead, one field will be named `Location` and the other field will be named after the `operationId`. Which operation will take on the `location` field name is not obvious. 

*** 

A possible solution is to produce field names form the operation path. This will allows us to create consistent naming for all Query fields. Mutation can follow a similar treatment but also append the operation method in the field name. 

However, if this makes assumptions that the URL can easy transformed into a sensible field name. The URL may contain a lot of meta-data concerning the version of the API or other redundant information and it will all appear in the GraphQL interface. In addition, the URL may be too short and nay not convey enough information about the purpose of the operation. 

[...]

Issue #131, opened 2019-03-17
Query type - irregular naming convention


dougal83:
One thing irks me about the current output of oasgraph is that we have query types with numbers added to the end.

Using this [example openapi schema](https://github.com/dougal83/express-lb4-ng/blob/feat/oasgraph/server/bin/openapi.json) from lb4.
### Current behaviour
We return a query such as 'Todos3'.
### Expected
A consistent naming convention.  'Todos'  Alternatively attaching the verb to avoid naming collisions? e.g: 'TodosGet'.  
### Discussion
I find that the [openapi-generator schema](https://openapi-generator.tech/docs/generators/graphql-schema) produces my idea of an expected naming convention.  For the above openapi schema we get the following: [graphql schema](https://github.com/dougal83/generated-graphql/tree/master/graphql).

Is this something that can be improved or unnecessary as you see it?

Alan-Cha:
@dougal83 Thanks for the feedback! I think appending the verb at the end of the type name is an option that we can add. Some users might find it useful. However, I do not think this will solve the core issue. 

What I mean is that even if we do append the verb at the end of the type name, in regards to some schemas, we will still need to add a number at the end as well, e.g. `TodosGet2` or `Todos2Get`. 

I believe that OASGraph does indeed have a consistent naming convention. Perhaps it is not immediately obvious and needs better documentation so I will try to give some insight here. 

OASGraph creates object types based on the operation's `preferredName` and the response schema. We run into trouble when two or more operations share the same `preferredName` but have different response schemas.

Although I have not taken a close look at the example openapi schema you have linked, I assume that in this schema, there are multiple operations that have the same `preferredName` and different response schemas, leading to operation types: `Todos`, `Todos2`, and `Todos3`. 

If all we do is append the verb to end of the `preferredName`, these collisions can still happen. Perhaps in this OAS, we will have nice names but in others, this problem can still be present. 

***

One way we can counteract this ugly naming is to use the [`operationIdFieldNames` option](https://github.com/strongloop/oasgraph/tree/master/packages/oasgraph#options). This allows a user to manually specify the object type names by changing the `operationId`s in the OAS. 

Alternatively, a user can try to collapse schemas with the same `preferredName`. If all operations with the same `preferredName` use the same response schema, then this problem will be solved. However, this could be quite involved and may lead to schemas with lots of optional fields. One potential route that we may take OASGraph is to do automatic schema collapsing, however we are still not so sure about that. 

***

[...]

PR #185, opened 2019-06-28
Infer fields names from path

ErikWittern:
Address #131, #179, #180, and #181

Signed-off-by: Erik Wittern <witternj@us.ibm.com>


packages/openapi-to-graphql/src/oas_3_tools.ts
```diff
@@ -38,7 +38,7 @@ import * as Swagger2OpenAPI from 'swagger2openapi'
 import * as OASValidator from 'oas-validator'
 import debug from 'debug'
 import { handleWarning } from './utils'
-import { parseType } from 'graphql'
+import * as pluralize from 'pluralize'
```
omarchehab98: Are we making an assumption that the API is in english? Perhaps we should expose an API to add pluralization rules from a JSON file

ErikWittern: Thank you for that comment! I think it was the implicit assumption so far that the API is in English, yes. But as you point out, there could be the need for pluralizing names for APIs in other languages.

Thinking about an option, we would propose to allow developers to optionally add their own `pluralize` function, which takes as input a string and is expected to output a (pluralized) string. The default `pluralize` library would be used if that option is not provided.

ErikWittern: A side comment: the string passed to `pluralize` is not necessarily a single word - it is basically the sanitized version of a path component, which could be anything, e.g., a combination of words.

wtrocki:
Awesome change! Let me know if I can help with review/testing.

Alan-Cha:
@wtrocki Thanks for the encouragement! I think, for the most part, this PR is ready to ship. We will adapt it slightly, following @omarchehab98's suggestion and expose the pluralization conversion to allow for other languages. 

The main road stop is we're trying to figure out when is a good time to release this change and how should we do it, as this will be a breaking change. Would this be a 2.0.0 release, for example?

packages/openapi-to-graphql/src/oas_3_tools.ts
ErikWittern: @omarchehab98 We wanna go ahead in the next days and merge this PR. We are thinking to not allow passing a custom `pluralize` for now, as we assume most interfaces to be in English (which is made likely by the constraints that the GraphQL spec has for names).

However, if we hear that people have specific needs for custom pluralize functions, we would reconsider. Do you actually have the need for a custom pluralize function?

omarchehab98: Some teams name things in other languages other than English, you can go without a custom pluralize function until someone opens an issue about it.

ErikWittern: Thanks for your response! I think this will be our way forward for now: not add the option until specific requirements are stated.

Alan-Cha:
Refer to #296

PR #186, opened 2019-06-28
Stringify objects with no properties

Alan-Cha:


Alan-Cha:
This is kind of a weird PR. I started out just trying to improve the error messages. After a few hours, I found myself here. 

Basically, a common problem we encounter is when OAS have an object schema that does not define an properties. This is troublesome because GraphQL has very strict typing requirements. This pull request will recursively stringify (sub-)objects that are missing properties.

packages/openapi-to-graphql/test/example_api_server.js
```diff
@@ -153,8 +153,8 @@ function startServer(PORT) {
 
   const Patents = {
     'CCC OSv1': {
-      patentId: '100',
```
Alan-Cha: Typo in the example API. It does not match the OAS. It is already sanitized here.

packages/openapi-to-graphql/test/example_api_server.js
```diff
@@ -479,7 +517,7 @@ function startServer(PORT) {
   app.get('/api/patents/:id', authMiddleware, (req, res) => {
     console.log(req.method, req.path)
     for (let patent in Patents) {
-      if (Patents[patent].patentId === req.params.id) {
```
Alan-Cha: Same thing as above.

packages/openapi-to-graphql/src/oas_3_tools.ts
```diff
@@ -411,17 +411,9 @@ export function getSchemaType(schema: SchemaObject): string | null {
     // CASE: arbitrary JSON
     if (typeof schema.additionalProperties === 'object') {
       return 'json'
+    } else {
+      return 'object'
     }
-
-    // If there are no properties:
```
Alan-Cha: I wanted to get rid of this. To return `null` in here is not only wrong but also ugly.

packages/openapi-to-graphql/src/preprocessor.ts
```diff
@@ -484,13 +481,10 @@ export function createDataDef(
     )
     const saneInputName = Oas3Tools.capitalize(saneName + 'Input')
 
-    // Add the names to the master list
-    data.usedOTNames.push(saneName)
```
Alan-Cha: We were adding `usedOTNames` here when we could not guarantee creating a dataDef. In other words, we potentially wasted the namespace here.

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -332,13 +331,14 @@ export function getResolver({
         if (err) {
           httpLog(err)
           reject(err)
-        } else if (response.statusCode > 299) {
+        } else if (response.statusCode < 200 || response.statusCode > 299) {
```
Alan-Cha: It should be like this, right? I wasn't sure because it seemed so simple...

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -364,13 +366,27 @@ export function getResolver({
...
               body = JSON.parse(body)
+            } else {
+              /**
+               * If not application/json, assume no additional processing is
+               * needed
+               *
+               * TODO: Should this also make a check to see what is defined in
+               * the schema? Probably not, would be an OAS/API mismatch.
+               */
+              resolve(body)
```
Alan-Cha: This is already covered by a test case FYI. We were supposed to resolve here but we didn't. Luckily, it didn't break anything before.

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -831,3 +847,41 @@ function graphQLErrorWithExtensions(
...
 }
+
+/**
+ * Ideally, all objects should be defined with properties. However, if they are
+ * not, then we can at least stringify the data.
+ *
+ * This function recursively ensures that objects have properties, if not
+ * stringify the data to match the mitigation.
+ */
+function stringifyObjectsWithNoProperties(
```
Alan-Cha: This is the main feature of this PR. Basically, ensure that the properties are defined, otherwise stringify.

PR #199, opened 2019-07-09
Improve error handling

Alan-Cha:
Mention keywords for combining schema


packages/openapi-to-graphql/src/utils.ts
```diff
@@ -13,9 +13,6 @@ export const mitigations = {
    * Should be caught by the module oas-validator
    */
   INVALID_OAS: `Ignore issue and continue.`,
-  INVALID_SCHEMA_TYPE: `Fall back to GraphQL string type and stringify returned data.`,
```
Alan-Cha: These errors are not used, are redundant, and/or are handled by other mitigation methods (like #186).

packages/openapi-to-graphql/src/preprocessor.ts
```diff
@@ -484,17 +502,11 @@ export function createDataDef(
     // Determine the type of the schema
     const type = Oas3Tools.getSchemaType(schema as SchemaObject)
     if (!type) {
-      // TODO: Throw error?
```
Alan-Cha: I think we originally had/were planning to have some stringify mitigation. Either way, that is redundant now because of #186.

Alan-Cha: As you can see from the comment, I already had doubts that this should be an error instead of a warning. I think making it a full error make sense. If we cannot understand the schema, then there is really no way to go forward. Or perhaps we can make a check as we are iterating through the operations to make sure that there is a response definition...

Alan-Cha: I don't think that strategy will work because the response definition could be nested. We need a check to see if a GraphQL object can be created.

PR #202, opened 2019-07-09
A few changes

Alan-Cha:
If the remote file was YAML, the CLI would not parse it correctly.


packages/openapi-to-graphql-cli/src/openapi-to-graphql.ts
```diff
@@ -141,7 +144,20 @@ function getRemoteFileSpec(uri) {
         } else if (res.statusCode !== 200) {
           reject(new Error(`Error: ${JSON.stringify(body)}`))
         } else {
-          resolve(body)
+          // If the body is an object, it is most likely an OAS in JSON format
+          if (typeof body === 'object') {
+            resolve(body)
+
+            // If the body is a string, try to parse as YAML
+          } else if (typeof body === 'string') {
```
Alan-Cha: I'm not too sure about this code... but I'm not sure what else we should do.

packages/openapi-to-graphql/test/example_api.test.js
```diff
@@ -130,7 +130,7 @@ test('Get resource with no response schema and status code: 204 and fillEmptyRes
   return graphql(createdSchema, query).then(result => {
     expect(result).toEqual({
       data: {
-        bonuses: ''
+        bonuses: null
```
Alan-Cha: Should fields that rely on fillEmptyResponses return null?

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -360,159 +367,191 @@ export function getResolver({
...
+            const {
+              responseContentType,
+              responseSchema
+            } = Oas3Tools.getResponseSchema(
+              operation,
+              operation.statusCode,
+              operation.oas
+            )
+            if (responseSchema === null) {
+              resolve(null)
```
Alan-Cha: See above

packages/openapi-to-graphql-cli/src/openapi-to-graphql.ts
```diff
@@ -131,7 +131,10 @@ function readFile(path) {
       : yaml.safeLoad(fs.readFileSync(path, 'utf8'))
     return doc
   } catch (e) {
-    console.error('Error: failed to parse YAML/JSON')
+    console.error(
+      `Failed to parse JSON/YAML. Ensure correct file ` +
+        `extension (i.e. '.json' or '.yaml').`
```
ErikWittern: I am unsure about this error message: we are testing for a `.json` ending, but the error message makes it seem that in addition, only `.yaml` is supported. I have seen quite a few `.yml` files for YAML also. Now, if something goes wrong parsing them, it may seem to users the reason is that the ending is `.yml` instead of `.yaml`, which is not the problem, though.

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -360,131 +367,191 @@ export function getResolver({
...
              */
-            if (response.headers['content-type'].includes('application/json')) {
-              body = JSON.parse(body)
+            if (
+              !(
+                response.headers['content-type'].includes(
+                  operation.responseContentType
+                ) ||
+                operation.responseContentType.includes(
+                  response.headers['content-type']
```
ErikWittern: Do we assume header keys are already normalized? E.g., what would happen if the header key is `Content-type`?

packages/openapi-to-graphql/test/example_api_server.js
```diff
@@ -416,6 +416,7 @@ function startServer(PORT) {
     console.log(req.method, req.path, req.query, req.headers)
     if ('cookie' in req.headers) {
       res
+        .set('Content-Type', 'text/plain')
```
ErikWittern: Above, we are looking for key `content-type`...

packages/openapi-to-graphql-cli/src/openapi-to-graphql.ts
Alan-Cha: Do we want to make a check for all possible endings (`.json`, `.yml`, and `.yaml`)?

packages/openapi-to-graphql/src/resolver_builder.ts
Alan-Cha: No, I do not think we can assume. See [here](https://github.com/request/request#responsecaselessgetheader-name). I will make the necessary changes.

Alan-Cha: Oh wait, I think this is good. I looked at the TS declarations.

Alan-Cha: All HTTP headers used in this way are lowercased.

packages/openapi-to-graphql-cli/src/openapi-to-graphql.ts
Alan-Cha: Do we want to make it a more stringent test? Instead of just an regex inclusion check, make sure that the last characters of the path are exact matches of the expected extensions?

packages/openapi-to-graphql-cli/src/openapi-to-graphql.ts
```diff
@@ -125,14 +121,15 @@ Promise.all(
  * @return {object}      Content of read file
  */
 function readFile(path) {
-  try {
```
Alan-Cha: Removed try-catch block for more streamline error handling... Not sure if this is a good practice...

packages/openapi-to-graphql-cli/src/openapi-to-graphql.ts
```diff
@@ -125,14 +121,15 @@ Promise.all(
...
-    const doc = /json$/.test(path)
-      ? JSON.parse(fs.readFileSync(path, 'utf8'))
-      : yaml.safeLoad(fs.readFileSync(path, 'utf8'))
-    return doc
-  } catch (e) {
-    console.error('Error: failed to parse YAML/JSON')
-    return null
+  if (/json$/.test(path)) {
```
Alan-Cha: Should these be exact match checks?

ErikWittern:
@Alan-Cha Looks good to me now.

PR #210, opened 2019-07-11
Tests in Typescript

Alan-Cha:

### DIFF
--- a/packages/openapi-to-graphql/test/authentication.test.ts
+++ b/packages/openapi-to-graphql/test/authentication.test.ts
@@ -7,10 +7,10 @@
 
 /* globals beforeAll, test, expect */
 
-const openapiToGraphql = require('../lib/index.js')
-const { graphql } = require('graphql')
+import * as openapiToGraphql from '../lib/index.js'
+import { graphql } from 'graphql'
 
-let oas = require('./fixtures/example_oas.json')
+const oas = require('./fixtures/example_oas.json')
 const PORT = 3003
 // update PORT for this test case:
 oas.servers[0].variables.port.default = String(PORT)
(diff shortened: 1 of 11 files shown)


Alan-Cha:
The servers (and the evaluation) are still in Javascript though. We can iterate on it.

packages/openapi-to-graphql/test/authentication.test.ts
```diff
@@ -147,12 +147,13 @@ test('Get project using API key passed as option - viewer is disabled', async ()
...
 test('Get project using API key passed in the requestOptions - viewer is disabled', async () => {
-  let { schema } = await openapiToGraphql.createGraphQlSchema(oas, {
+  const { schema } = await openapiToGraphql.createGraphQlSchema(oas, {
     viewer: false,
     requestOptions: {
       headers: {
         access_token: 'abcdef'
-      }
+      },
+      url: undefined
```
Alan-Cha: This is kind of ugly... not sure if there's much we can do about it. We rely on the request module for this object type definition.

packages/openapi-to-graphql/test/example_api.test.ts
```diff
@@ -841,12 +842,12 @@ test('Capitalized enum values can be returned', () => {
...
 
 test('Define header and query options', () => {
-  let options = {
+  const options = {
     headers: {
       exampleHeader: 'some-value'
     },
     qs: {
-      limit: 30
+      limit: '30'
```
Alan-Cha: Typo caught because of Typescript

ErikWittern: Nice!

packages/openapi-to-graphql/test/authentication.test.ts
ErikWittern: That's an ok price to pay, IMO.

ErikWittern:
@Alan-Cha Nice, I am super happy we are using TypeScript in tests now, because it (likely) more closely emulates how OtG is used.

PR #221, opened 2019-08-08
Add additional options to the CLI

Alan-Cha:


ErikWittern:
@Alan-Cha Very nice work!

Issue #67, opened 2018-11-05
Support YAML


Alan-Cha:
According to the [specification](https://swagger.io/specification/): "An OpenAPI document that conforms to the OpenAPI Specification is itself a JSON object, which may be represented either in JSON or YAML format."

We currently only support JSON. We should make a check for YAML. 

Related: issue #66

ErikWittern:
@Alan-Cha I think that we should consider this in relation to separating OASGraph core from OASGraph CLI (https://github.com/strongloop/oasgraph/issues/59). I don't think that the core component requires YAML support, because in the use as a library, the application using OASGraph is responsible for producing an in-memory representation of the OAS. The CLI, however, should be able to handle YAML documents for convenience purposes.

wtrocki:
I came over this. I fully agree with the author. YAML support is not needed, however it will be nice to have this in the readme as people may still try to use YAML examples. Creating quick PR.

Alan-Cha:
@wtrocki Thanks for your input! I believe there may be some miscommunication. The CLI has JSON/YAML support and the library only has JSON support. 

The PR you filed #236 is a bit misleading as the CLI does indeed have YAML support.

wtrocki:
@Alan-Cha - ok. I was using sample yaml file and got strange result so assumed that it did not supported it. I will close PR. Thank you for very quick response.

Alan-Cha:
@wtrocki Oh, interesting! Were you able to diagnose the problem?

wtrocki:
I got now example petstore api and it simply works so problem was 100% on my side. I started messing with extra extension fields for https://github.com/IBM/openapi-to-graphql/issues/235

PR #236, opened 2019-09-02
Add note that CLI supports only json file 

wtrocki:
Add note that CLI supports only json file.

Signed-off-by: Wojciech Trocki wtr@redhat.com


Alan-Cha:
@wtrocki Thanks for the PR! As I mentioned in #67, I think there may be a misunderstanding. The CLI does indeed have YAML support.

PR #247, opened 2019-09-09
Improvements for generated schema string 

wtrocki:
## Motivation

When generating schema to file, graphql schema contains a lots of comments. 
Because of the default formatting each comment has additional line break, making entire generated schema very long and hard to read. With more advanced oas we can easily get ~700 lines looking like this:

```

  """No description available."""
  isUKApproved: Boolean

  """No description available."""
  mandatoryCustomerReference: String

```

After this fix schema can look like this

```
isUKApproved: Boolean
mandatoryCustomerReference: String
```

Creating an idea. Feel free to close it.


Alan-Cha:
@wtrocki I think this is a good change! It takes a fresh pair of eyes to find problems like this! I made some changes, but once the checks finish, I will pull it in.

Issue #177, opened 2019-06-25
Enable (HTTP) logging per default in CLI


ErikWittern:
OpenAPI-to-GraphQL logs based on the value of the `DEBUG` environment variable, allowing users to [control what is being logged](https://github.com/IBM/openapi-to-graphql/tree/master/packages/openapi-to-graphql#logging).

IMO, the CLI should per default enable logging, as it makes it easier for users to understand how the library works and eventually debug things.

As part of enabling logging per default, the CLI could make logging controllable using an option.

PR #249, opened 2019-09-17
Debug support

wtrocki:
As in https://github.com/IBM/openapi-to-graphql/issues/177#event-2640312411


Alan-Cha:
@wtrocki Thank you so much for this contribution! And sorry for the late reply, it's been especially busy in the office. This is a great change. I will pull it in later today.

wtrocki:
Just tested this and there is a side impact of this change. If any dependency have debug it will print it as well. To resolve that we could add prefix for library and then use it instead of `*`

Alan-Cha:
Wow! Good call. I think we could arguably just assign it to `preprocessing,translation,http` as those are the only [logging options](https://github.com/IBM/openapi-to-graphql/tree/master/packages/openapi-to-graphql) we offer as of now.

wtrocki:
Thank you. IMHO `otg:` prefix will be better for future maintenance but lets do it this way first!

wtrocki:
Unrelated question. Should I be committing js files as well?

Alan-Cha:
@wtrocki Yes, the JS files should be committed as well.

wtrocki:
Ok let me push changes again then

Alan-Cha:
@wtrocki No need! I'm actually making a bunch of changes on your branch. I can take over now.

wtrocki:
Ups.. sorry - feel free to remove things

Alan-Cha:
@wtrocki No problem! Thanks!

Alan-Cha:
@wtrocki I am a bit confused... it seems like we aren't getting the debug messages.

***

It seems like...

```bash
cd openapi-to-graphql/packages/openapi-to-graphql-cli

DEBUG=preprocessing,translation,http node lib/openapi-to-graphql.js ../openapi-to-graphql/test/fixtures/example_oas.json 
```

...and...

```bash
cd openapi-to-graphql/packages/openapi-to-graphql-cli

node lib/openapi-to-graphql.js ../openapi-to-graphql/test/fixtures/example_oas.json 
```

...have very different behaviors.

***

I took a look around and it seems like we're doing all the right things but we're not getting the expected output. Will dig around a bit further...

Alan-Cha:
I think I know what the problem is. We create the debug loggers in the global space ([for example](https://github.com/IBM/openapi-to-graphql/blob/8a41f9c749e119b152a48edb514b870b28c484f4/packages/openapi-to-graphql/src/index.ts#L59)) and we import OtG at the [top of the file](https://github.com/IBM/openapi-to-graphql/blob/8a41f9c749e119b152a48edb514b870b28c484f4/packages/openapi-to-graphql-cli/src/openapi-to-graphql.ts#L13), before we process the CLI options. Basically, we have already processed the DEBUG environment variables before we have the chance to set them from the CLI options. 

I'm not sure what the best way to move forward is. It seems to me we have two options, instantiate the debug loggers elsewhere and pass them around (perhaps through the `data` object) or move the import of OtG later in the code. Both aren't great. The second seems to be bad coding practice as well. 

Thoughts? @wtrocki

wtrocki:
Right. Option one will be to put environment variables in CLI without the ability to change them before import:

https://github.com/IBM/openapi-to-graphql/blob/8a41f9c749e119b152a48edb514b870b28c484f4/packages/openapi-to-graphql-cli/src/openapi-to-graphql.ts#L12-L13

There is no other option that will be sensible or require serious refactoring.

wtrocki:
Happy to close this if we think that this does not make sense anymore.

Alan-Cha:
@wtrocki I think it won't be that much refactoring. 

So let's say we just instantiate all of the debug loggers at the beginning of `translateOpenApiToGraphQL()` and stuff them into the preprocessing data, do you think that makes sense?

```Javascript
data['loggers'] = {
  httpLog: debug('http')
  preprocessingLog: debug('preprocessing')
  translationLog: debug('translation') 
}
```

Somewhere around [here](https://github.com/IBM/openapi-to-graphql/blob/8a41f9c749e119b152a48edb514b870b28c484f4/packages/openapi-to-graphql/src/index.ts#L209)...

*** 

Thoughts?

wtrocki:
I remember using `debug.enable('*')` Need to check if debug have some global enabler

packages/openapi-to-graphql/src/index.ts
```diff
@@ -56,15 +55,21 @@ type Result = {
...
-const translationLog = debug('translation')
-
 /**
  * Creates a GraphQL interface from the given OpenAPI Specification (2 or 3).
  */
 export async function createGraphQlSchema(
   spec: Oas3 | Oas2 | (Oas3 | Oas2)[],
   options?: Options
 ): Promise<Result> {
+  const debug = require('debug') // Cannot import globally because it depends on the current state of the DEBUG environment variable
```
Alan-Cha: This is pretty bad...

Alan-Cha:
@wtrocki I was able to make the necessary changes to get the intended result, basically instantiate the debuggers and pass them around. However, to my dismay, I still needed to move the debug module import. 

If enabling will be able to fix everything in a simple and unintrusive way, then let's give it a go. Can you give it a shot? If so, could you submit another PR?

packages/openapi-to-graphql/src/index.ts
Alan-Cha: We also can't use TS typing because of this pattern.

PR #255, opened 2019-10-03
Add publish registry

Alan-Cha:
For https://github.com/IBM/openapi-to-graphql/packages

Not sure if this is a good idea? Doesn't seem to hurt...


wtrocki:
Currently they support only scoped packages so it will be hard to migrate without rename.

wtrocki:
https://help.github.com/en/articles/configuring-npm-for-use-with-github-package-registry#publishing-a-package

PR #264, opened 2019-10-24
Fix null value in request body

### LINKED ISSUE #263: Null values in request body are replaced with empty objects, preventing POST requests from being called properly
Hi @Alan-Cha or anyone,

Sample "POST" mutation with payload:
```
mutation {
  sampleMutation(requestBodyInput: [{name: "Yoyo", price: null}, {name: "Apple", price: null}])
}
```

From logging an HTTP request => `request` package:
```
{
    "json": true,
    "jsonReplacer": [Function: jsonReplacer],
    "method": "post",
    "url": "<URL>",
    "headers": {
      "content-type": "application/json",
      "accept": "application/json",
      "Authorization": "<AUTH TOKEN>"
    },
    "body": "[{ 'name: 'Yoyo', price: {} }, { 'name': 'Apple', price: {} }]"
  }
```

The replacing of the price values from `null` to `{}` is causing my POST request to fail. I attempted to place a jsonReplacer function inside the `requestOptions` to replace the empty objects back to `null` before sending the request, but no luck there. Any quick tips on how I can preserve the `null` values, or get the jsonReplacer function to properly pass? I have no control over this API that I am attempting to hit FYI. Also, I assume the `null` values are being replaced because of something to do with GraphQL?

```
requestOptions: {
  json: true,
  jsonReplacer: (key, value) => {
    if (Object.entries(value).length === 0 && value.constructor === Object) return null
    return value
  },
},
```

Alan-Cha:
Fix #263


ErikWittern:
@Alan-Cha Looks good to me. Once you rebase, I can merge.

PR #269, opened 2019-11-12
Add option for generic payload argument name, 'requestBody', for mutations

yoonslee:
Currently, POST endpoints that get transformed to the GraphQL mutations have a sanitized argument name for request bodies, e.g., `catsEditV1` from `POST /cats/edit/v1`. I propose a more generic term for the request body: `requestBody`, which is used in OAS.

Signed-off-by: Yoon Lee <himynameisyoon@gmail.com>


Alan-Cha:
@yoonslee This is awesome!!! Thank you so much!!! I would like to add a test case, which I am happy to do, but otherwise, this looks good to go from me!

yoonslee:
@Alan-Cha Awesome! I also notice a typo in README.md:159 = `Otherwise, the name will default [to] the camelCased pathname.` Missing [to]. Going to fix that!

yoonslee:
@Alan-Cha Done, updated! Let me know if there's anything else I can do. I do wonder how to best write a test for this.

Alan-Cha:
@yoonslee There are a few options. 

One way is to [generate a new schema](https://github.com/IBM/openapi-to-graphql/blob/a8203694adbf26e78f1f64bb3d1991a34c1b8094/packages/openapi-to-graphql/test/example_api.test.ts#L1048) using the new option like [this](https://github.com/IBM/openapi-to-graphql/blob/a8203694adbf26e78f1f64bb3d1991a34c1b8094/packages/openapi-to-graphql/test/example_api.test.ts#L1039), then running a mutation query like [this](https://github.com/IBM/openapi-to-graphql/blob/a8203694adbf26e78f1f64bb3d1991a34c1b8094/packages/openapi-to-graphql/test/example_api.test.ts#L690) but with the correct argument name, and then verifying the response. 

Alternatively, we can do it by introspecting the query like [this](https://github.com/IBM/openapi-to-graphql/blob/a8203694adbf26e78f1f64bb3d1991a34c1b8094/packages/openapi-to-graphql/test/example_api.test.ts#L1676).

If you'd like to take a shot at it, be my guest! That would be really helpful as I'm bogged down working on a different PR. I'm sorry I haven't got around to adding the test yet 😅

packages/openapi-to-graphql/lib/types/options.d.ts
```diff
@@ -184,6 +188,10 @@ export declare type InternalOptions = {
      * users to specify other formats that should be interpreted as ID types.
      */
     idFormats?: string[];
+    /**
+     * Custom headers to send with every request made by a resolve function.
```
Alan-Cha: Small error here. Seems like the wrong comment...

Alan-Cha: Ah it just needs to be rebuilt.

Alan-Cha: Done!

yoonslee:
@Alan-Cha I got this! Will definitely work on the test over the next few days.

Alan-Cha:
@yoonslee I finally added those tests! I think we're ready to pull in! Again, thanks so much for trying to contribute to this project!!!

packages/openapi-to-graphql-cli/README.md
```diff
@@ -33,6 +33,7 @@ Options:
   -o, --operationIdFieldNames    create field names based on the operationId
   -f, --fillEmptyResponses       create placeholder schemas for operations with no response body rather than ignore them
   -a, --addLimitArgument         add a limit argument on fields returning lists of objects/lists to control the data size
+  -g, --genericPayloadArgName    Sets argument name for the payload of a mutation to 'requestBody'
```
ErikWittern: Super-nit-picky: maybe we want to say `of any mutation` here?

Issue #234, opened 2019-08-31
Pass headers to backend


YifengWong:
Instead of using apikey viewer things, I want to pass my cookie each request in header to backend, just pass ALL Cookie, any resolution? thx

Alan-Cha:
Hi! Sorry for the late response. I believe this is possible using the `requestOptions` (to pass the cookie) and `viewer` (to remove viewers) [options](https://github.com/IBM/openapi-to-graphql/tree/master/packages/openapi-to-graphql#options).

***

See an example [here](https://github.com/IBM/openapi-to-graphql/blob/0ecb7f7573db0c6931ad5d7e12b59d94db7e521b/packages/openapi-to-graphql/test/authentication.test.ts#L301).

```
const { schema } = await openapiToGraphql.createGraphQlSchema(oas, {
  viewer: false,
  requestOptions: {
    headers: {
      cookie: 'access_token=abcdef'
    },
    url: undefined
  }
})
```

***

The `requestOptions` directly exposes the [Request.js](https://www.npmjs.com/package/request) options used by the resolvers to call the REST API. You can find other options [here](https://github.com/request/request#requestoptions-callback).

YifengWong:
@Alan-Cha Thanks for your response. In fact I want to pass different header in each request to backend, I have read the source code, it seems that it is unachievable.

Is it possible to add a requestOptions to allow user to transfer some fields in header for each request? such as

`
requestOptions: {
  penetrateHeader: ['cookie']
}
`

in resolver builder:
`
if (penetrateHeader.length > 0) {
	penetrateHeader.forEach(fieldName => {
		options.headers[fieldName] = ctx.headers[fieldName]
	})
}
`

Alan-Cha:
@YifengWong Happy to help! So you want the user to be able to change the header in each request? If so, it should just be a matter of defining header parameters in the OAS. Then, OtG will automatically create an interface such that each operation (requiring header parameters) will have additional arguments that the user can use for header parameters. 

Is this what you mean?

***

[Here](https://github.com/IBM/openapi-to-graphql/blob/f44205a9035ac2db08cc6ebf9d499dc67520e0fc/packages/openapi-to-graphql/test/example_api.test.ts#L569) is an example of a query that has arguments for passing cookie values.

[...]

PR #270, opened 2019-11-12
Allow 'headers' resolver option to be a function

yoonslee:
Signed-off-by: Yoon Lee <himynameisyoon@gmail.com>

### DIFF
--- a/packages/openapi-to-graphql/README.md
+++ b/packages/openapi-to-graphql/README.md
@@ -160,7 +160,7 @@ Schema options:
 
 Resolver options:
 
-- `headers` (type: `object`, default: `{}`): Headers to be sent in every request to the API described by the given OAS. Parameters defined in the OpenAPI Specification to set these headers will be ignored by OpenAPI-to-GraphQL.
+- `headers` (type: `object` | `function`, default: `{}`): Headers to be sent in every request to the API described by the given OAS. Parameters defined in the OpenAPI Specification to set these headers will be ignored by OpenAPI-to-GraphQL. If used as a function, the following parameters will be exposed per-request: `req` (the GraphQL resolver's context.request object), `method`, `path`, and `title`.
 
 - `qs` (type: `object`, default: `{}`): Query parameters to be sent in every request to the API described by the given OAS. Parameters defined in the OpenAPI Specification to set these query parameters will be ignored by OpenAPI-to-GraphQL.
 
--- a/packages/openapi-to-graphql/src/resolver_builder.ts
+++ b/packages/openapi-to-graphql/src/resolver_builder.ts
@@ -13,6 +13,7 @@ import { Operation, DataDefinition } from './types/operation'
 import { ResolveFunction } from './types/graphql'
 import { PreprocessingData } from './types/preprocessing_data'
 import * as NodeRequest from 'request'
+import { RequestHeadersFunction } from './types/options'
 
 // Imports:
 import * as Oas3Tools from './oas_3_tools'
@@ -36,13 +37,17 @@ type AuthOptions = {
   authCookie: NodeRequest.Cookie
 }
 
+type RequestOptions = Omit<NodeRequest.OptionsWithUrl, 'headers'> & {
+  headers: { [key: string]: string } | RequestHeadersFunction
+}
+
 type GetResolverParams = {
   operation: Operation
   argsFromLink?: { [key: string]: string }
   payloadName?: string
   data: PreprocessingData
   baseUrl?: string
-  requestOptions?: NodeRequest.OptionsWithUrl
+  requestOptions?: RequestOptions
 }
 
 /**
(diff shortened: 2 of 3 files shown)


Alan-Cha:
@yoonslee Thank you for the pull request! Although, I don't exactly understand how this change will improve the project, at least not in its current state. The function does not take any arguments, so how is it any different from providing the headers directly, as an object? I would see why making the `headers` option a function could be helpful, by dynamically setting the headers, but it does not seem capable of doing that right now. Could you clarify?

yoonslee:
@Alan-Cha Thanks for the follow-up! Looks like I applied the wrong stash, and I see that the change do not actually improve or work as intended. As mentioned in issue #234, the idea was to expose the context on a per-request basis so that I could pass through custom headers to the outbound request or in another POV, persist the headers that a request originally came with, even as it leaves. This is especially useful for microservices that sit between two backends. The header that I am trying to pass is an API key, but I would prefer not to use the `viewerApiKey` wrapper as it adds an extra layer to the generated schema. Would love to hear your thoughts on this.

Alan-Cha:
@yoonslee Yes, I recall! I think what we need to do is to pass some subset of the data contained in [`operation`](https://github.com/IBM/openapi-to-graphql/blob/a8203694adbf26e78f1f64bb3d1991a34c1b8094/packages/openapi-to-graphql/src/resolver_builder.ts#L53). At the very minimum, it should the path, the method, and the OAS (title) of the operation. I'm not sure what else we should provide though. We could expose everything the resolver will get (i.e. `root`, `args`, `ctx`, and `info`) but I wonder if that's overkill. It may also be a slippery slope. 

Thoughts?

yoonslee:
First, thank you! I definitely appreciate the feedback and interaction.

How about...
If the `requestOptions.headers` option is used as a function, the following parameters will be exposed per-request: `req` (the GraphQL resolver's context.request object), `method`, `path`, and `title`.

Since this is an option inside `requestOptions`, I think semantically it makes sense to only expose the request object inside resolver's context. I do think that the others, `args`, `info`, etc., are not as essential (at least for my use case).

Anyhoo, I pushed up a new version of this branch, and would like to see this or something equivalent be a part of this project!

How I see myself using it:
I would need to expose the context in the GraphQL server for any incoming requests. Then, it would be as easy as passing along whatever I want. In this example, the `Authorization` token header.
```
  const { schema } = await createGraphQlSchema(oas, {
    requestOptions: {
      headers: ({ path, method, title, req }) => console.log({ path, method, title }) || ({
        Authorization: req.headers.Authorization,
      }),
    },
  })

  const server = new GraphQLServer({
    schema,
    context: (ctx) => ctx,
  })
```

Alan-Cha:
@yoonslee Thank you so much for you hard work! Again, this is all super helpful. 

I think we are moving in the right direction but there are a few things I would like to clarify. There are two ways to inject custom headers and that is through the options `headers` and `requestOptions` (specifically `requestOptions.headers`). They completely overlap each other so we should have considered removing one of them a long time ago.

In this PR, you have changed the description for `headers`, removed the functionality for `headers`, and modified the functionality for `requestOptions`. I just wanted to make sure you are aware. 

One of the issues I foresee is that in the PR, we are using a custom type for `requestOptions` rather than relying on the one provided by the request module. This isn't necessarily an issue in and of itself but I think that it was a nice feature that we used and exposed the request-provided type. I am not strictly against using a custom type but I just want the repercussions to be considered. How do we convey that this is a superset of the request-provided type to the user. Is this a slippery slope? With the current changes, an object of the request-provided type can be assigned to the `RequestOptions` type, but if we continue to allow changes to the type, will there a point where this relationship is untrue? And what do we do in that case? 

That is why I made this [comment](https://github.com/IBM/openapi-to-graphql/issues/234#issuecomment-541830575). 

My personal opinion is that we should only have one way to inject custom header and that is through `requestOptions`. I am also okay with using a new custom type, as this will allow to add new and useful features in the future. 

I would like @ErikWittern and @wtrocki to weigh in if possible.

mikestaub:
I had to unblock my team so I forged and merged a PR based on this one, let me know if you want me to submit the PR here as well: https://github.com/mikestaub/openapi-to-graphql/pull/3

isocroft:
Has this PR been merged ?

Alan-Cha:
@isocroft Ahhh I feel very bad for letting this slip by for so long. No, this has not been merged yet. I began reworking this yesterday night, trying to add new tests, but I discovered problems with existing tests. I might not be able to finish this end of today because of other work I need to complete but I will have a call with my colleague and we should at least get some of the other PRs checked in today.

isocroft:
@Alan-Cha can we take a look at [these changes](https://github.com/mikestaub/openapi-to-graphql/pull/3/files) and compare them to the ones submitted here and figure out which diffs makes the existing tests pass and which do not. 

After analyses, I feel the changes to [this line of code](https://github.com/mikestaub/openapi-to-graphql/pull/3/files#diff-0cc951405b58dc28d896fdbf336d82bfR139) is what is making the existing tests not to pass

S1019269:
Hello guys, following the discussion here. We are having the very same usecase and needs of propagating the headers to the backend as @yoonslee mentions in the comments. I was wondering if you expect this PR to be merged in the near future.

Alan-Cha:
@S1019269 A while ago, I mentioned I was reworking the PR. I was able to identify the problems I was seeing in the existing test cases and address them, however, I got distracted and started working on other lines of work. I believe the PR is nearly complete, just needing some new test cases to test the new feature. I have some time today and I'll see if I can get it in.

Alan-Cha:
@S1019269 I created a PR with the updated changes but for some reason, I'm having a lot of trouble getting Travis to run the tests. The tests run fine on my local environment but for some reason Travis is having trouble downloading the modules. I will have to look into this tomorrow.

Alan-Cha:
I just accepted #329 and published version v2.2.0! This feature is now in! Sorry for keeping you all waiting but this long awaited feature is finally in. Thank you for everyone who gave me your feedback and help!!!

Issue #274, opened 2019-11-23
Problematic desanitization


Alan-Cha:
Drawn from #266 

As you may be aware of, OtG sanitizes strings from the OAS and then reverses the process where necessary in order to make the appropriate call. Unfortunately, some users have begun reporting problems because the method we use to do this is not waterproof. 

The method we currently use is, honestly, a bit primitive. The summarize, we have a map named saneMap that correlates sanitized strings with their raw versions. There are a few places where we store strings but most of time, it happens here, where we recursively analyze the different schemas that are used in the OAS.

This current method is a problem because we arbitrarily dumping all of the sanitized and raw strings into the same place without making any effort to consider their contexts. Therefore, when we actually try to desanitize the strings here, we encounter problems.

We went with this method because the "right" way to do it is complex and this was a more reasonable solution during our prototyping phase. I think the right way to do it would be to create a map for GraphQL (input) object type names, and multiple maps for field names in each of those GraphQL (input) object types.

However, I'm not sure if I have the time to make all the changes needed. If you would like to help, or anyone else, I would be happy to assist. I think for now, we can at least create two separate maps for GraphQL (input) object type names and one for field names. I believe this strategy will solve a significant percentage of possible conflicts by itself.

Issue #273, opened 2019-11-22
enum values remove underscores


mikestaub:
The spec we are wrapping has enum fields that have underscores, but the generated graphQL schema has them removed. This is causing the queries to break, is there a way to disable this behavior?

Alan-Cha:
@mikestaub Thanks for filling this issue. This must be a bug. I know you reached out earlier about another issue regarding enums and even created a sample repository. Unfortunately, I haven't had a chance to look at that issue either. Let me take a look now and see if I can give you more insight into the problem. Thanks for your patience.

Alan-Cha:
@mikestaub I was able to follow up on #266 and I wonder if the issues are related. Could you let me know if these enum fields correspond to other GraphQL components? I wonder if this is related to the problematic desanitization that I mentioned in the other issue.

mikestaub:
These enum fields are specific to just one type and are not shared. I don't think the issues are related.

Alan-Cha:
@mikestaub Is it possible to share the OAS (or a portion of it) so I can try to debug?

mikestaub:
Yes, here is a simplified example: 
```
swagger: "2.0"
info:
  version: "1.0.0"
  title: Test API

host: localhost:1339
basePath: /

schemes:
  - http
  - https

consumes:
  - application/json
produces:
  - application/json

x-servers:
  - url: //api.test.com
    description: Default server

paths:
  /v1/files:
    post:
      summary: Create a file
      operationId: postFile
      description: Creates a new file
      responses:
        "201":
          description: Success
          schema:
            $ref: "#/definitions/File"

definitions:
  Type:
    enum:
      - TEST_ENUM
    type: string

  File:
    type: object
    properties:
      cadType:
        $ref: "#/definitions/Type"

```

Alan-Cha:
I working on a fix #277. I have the desired behavior but it is breaking a few other tests which I'm still trying to figure out.

Issue #98, opened 2019-01-29
Support GraphQL conventions


Alan-Cha:
Conventions are important because they allow developers to create projects that are consistent and easily understandable by other developers. 

What are some of these conventions and how can we make sure OASGraph supports them?

ErikWittern:
Just for completeness, some are:

* PascalCased type names (see #99)
* camelCased field names (OASGraph already does this)
* CAPITALIZED enum values
* `Input` postfix for Input Object Types (OASGraph already does this)

Alan-Cha:
I see, so the next step would be to convert enum values to all-caps. Should this be an option?

ErikWittern:
W.r.t. enum values: If we change a value, let's say from `hour` to `HOUR`, we need to make sure that at runtime, we translate back and forth. I think we should try to stick to GraphQL conventions as much as possible, so I propose to default to making enum values all caps generally. I don't see why there needs to be an option to disable this behavior - in the end, the GraphQL interface has its own UX, which is going to be different from the REST(-like) API.

JennieJi:
Would you also transform all the REST API response fields to make sure resolvers work as expected (I met a bug that my API responded in lowercase and underscore separated, but Graphql transformed all these to camelcase)?

Alan-Cha:
We currently have support for GraphQL conventions with this [commit](https://github.com/IBM/openapi-to-graphql/commit/bfdc2f1ff9288561474eb7574c7dc211c847c364) and the release of v2.0.0

PR #277, opened 2019-11-25
Enum ALL_CAPS

Alan-Cha:
Work in progress


Alan-Cha:
Support for ALL_CAPS enum values

packages/openapi-to-graphql/lib/oas_3_tools.d.ts
```diff
@@ -156,15 +156,20 @@ export declare function getSecuritySchemes(oas: Oas3): {
...
+export declare enum CaseStyle {
+    PascalCase = 0,
+    camelCase = 1,
+    ALL_CAPS = 2
+}
 /**
  * First sanitizes given string and then also camel-cases it.
  */
-export declare function sanitize(str: string, lowercaseFirstChar?: boolean): string;
+export declare function sanitize(str: string, caseStyle: CaseStyle): string;
```
Alan-Cha: Sanitize now takes in a case. Before we were changing the case everywhere it was not very clear.

packages/openapi-to-graphql/src/oas_3_tools.ts
```diff
@@ -1026,64 +1026,78 @@ export function getSecurityRequirements(
...
+}
+
 /**
  * First sanitizes given string and then also camel-cases it.
  */
-export function sanitize(
-  str: string,
-  lowercaseFirstChar: boolean = true
-): string {
+export function sanitize(str: string, caseStyle: CaseStyle): string {
```
Alan-Cha: sanitize() now takes in a case. Before we were changing the case everywhere it was not very clear.

packages/openapi-to-graphql/src/oas_3_tools.ts
```diff
@@ -1026,64 +1026,78 @@ export function getSecurityRequirements(
   return results
 }
 
+export enum CaseStyle {
```
Alan-Cha: These are the different cases that we support. I wonder if I should keep the enum values themselves consistent instead of trying to replicate the cases.

packages/openapi-to-graphql/src/oas_3_tools.ts
```diff
@@ -1026,64 +1026,78 @@ export function getSecurityRequirements(
...
-
   return sanitized
 }
 
 /**
  * Sanitizes the given string and stores the sanitized-to-original mapping in
  * the given mapping.
  */
-export function sanitizeAndStore(
+export function storeSaneName(
```
Alan-Cha: This didn't really make sense trying to squash two different operations into the same function.

I think this is also a needed step for #274

packages/openapi-to-graphql/src/types/operation.ts
```diff
@@ -41,16 +50,23 @@ export type DataDefinition = {
    */
   subDefinitions: DataDefinition | { [fieldName: string]: DataDefinition }
 
-  links: { [key: string]: LinkObject }
+  // GraphQL-related:
+
+  // The potential name of the GraphQL type if it is created
+  graphQLTypeName: string
```
Alan-Cha: Calling this `otName` was a misnomer as the type we created could have been multiple different types, specifically object, array, or enum. 

I think this could also be a bit confusing as this is just a potential name, not the exact GraphQL type name, but I think it is an improvement over what we had before. Same for the other changes below.

packages/openapi-to-graphql/src/types/preprocessing_data.ts
```diff
@@ -42,7 +42,7 @@ export type PreprocessingData = {
   /**
    * List of all the used object names to avoid collision
    */
-  usedOTNames: string[]
+  usedTypeNames: string[]
```
Alan-Cha: This contained not only object type names but also array, enum, and input object type names

packages/openapi-to-graphql/test/example_api.test.ts
```diff
@@ -92,7 +92,7 @@ test('Get resource (incl. enum)', () => {
   }`
   return graphql(createdSchema, query).then(result => {
     expect(result).toEqual({
-      data: { user: { name: 'Arlene L McMahon', status: 'staff' } }
+      data: { user: { name: 'Arlene L McMahon', status: 'STAFF' } }
```
Alan-Cha: Enums are now in ALL_CAPS.

packages/openapi-to-graphql/src/schema_builder.ts
```diff
@@ -368,27 +383,27 @@ function createOrReuseEnum({
...
   } else {
-    translationLog(`Create GraphQLEnumType '${def.otName}'`)
+    translationLog(`Create GraphQLEnumType '${def.graphQLTypeName}'`)
 
     const values = {}
     def.schema.enum.forEach(e => {
-      // Force enum values to string
-      values[Oas3Tools.sanitize(e.toString(), false)] = {
+      // Force enum values to string and value should be in ALL_CAPS
+      values[Oas3Tools.sanitize(e.toString(), Oas3Tools.CaseStyle.ALL_CAPS)] = {
```
Alan-Cha: Changing enums to ALL_CAPS.

mikestaub:
Is there anything I can do to help get this merged?

Alan-Cha:
@mikestaub Thanks for the follow up! I think for the most part, this is ready to go. I am just waiting for @ErikWittern to review the PR and give me the final thumbs up. Feel free to review the code as well if you have the time. Unfortunately, I believe Erik is out this week... I would consider accepting the PR if it's a much needed fix/feature but because this will be a breaking change, I would like to go through the process. Opinions?

mikestaub:
Let's wait for Erik to take a look, it is a blocker for us, but it can wait a few more days.

Alan-Cha:
@mikestaub Unfortunately, still haven't heard word from @ErikWittern. Besides that, I think there is another problem on our hands. After taking a second look at this PR, I do not think it correctly solves the issue. It's actually more in line with #98 than #273. 

This PR changes enum values to ALL_CAPS rather than try to preserve the raw enum values. 

I don't know what I was thinking and I ended up solving a different problem.

I think what we need to do is avoid sanitizing the enum values and use the raw enum values [here](https://github.com/IBM/openapi-to-graphql/blob/ebaf00fc9b2e9cae227f6f3eb1acf2acac747dbc/packages/openapi-to-graphql/src/schema_builder.ts#L388). However, I am not sure if that's all we need to do as there is logic to desanitize the enum values as well. 

I'm so sorry for this. I can take another look at this problem tomorrow.

mikestaub:
I see. This PR actually solves my specific issue as we are already using all caps in our enums, we just have underscores. As long as you unsanitize the all caps back to raw values when making the actual request calls, I think it is fine. Using all caps for enums is a common convention in GraphQL schemas.

Alan-Cha:
In that case, because it solves your problem and because I'm not sure when Erik will have the time to review this, I can add this in as an option for now. That way, it will not be a breaking change. 

However, moving forward, I still think we should give users the ability to use the raw enum values. The question is what should be the default behavior, using the raw enum values, which users may be more familiar with, or using ALL_CAPS enum values, which follows GraphQL convention.

mikestaub:
I think the default should be to follow graphQL convention, as long as we allow the consumer to use the raw enums if they wish.

packages/openapi-to-graphql/src/types/operation.ts
ErikWittern: Agreed.

packages/openapi-to-graphql/src/types/preprocessing_data.ts
ErikWittern: Makes sense.

packages/openapi-to-graphql/src/oas_3_tools.ts
ErikWittern: It's a bit quirky trying to emulate the cases, but drives home the point what they are.

packages/openapi-to-graphql/src/oas_3_tools.ts
ErikWittern: I like that.

packages/openapi-to-graphql/src/oas_3_tools.ts
ErikWittern: Jupp, I like separating these two capabilities.

packages/openapi-to-graphql/src/preprocessor.ts
```diff
@@ -470,41 +481,53 @@ export function createDataDef(
...
-    const name = getSchemaName(data.usedOTNames, names)
+    const name = getSchemaName(names, data.usedTypeNames)
 
-    // Store and sanitize the name
-    const saneName = Oas3Tools.capitalize(
-      Oas3Tools.sanitizeAndStore(name, data.saneMap)
+    /**
+     * Store and sanitize the name
+     *
+     * TODO: Fix saneName store to avoid using camelCase and capitalizing it
```
ErikWittern: Not sure I understand the TODO. What is the issue here?

packages/openapi-to-graphql/src/oas_3_tools.ts
```diff
@@ -1026,64 +1026,74 @@ export function getSecurityRequirements(
...
+      // The first character in PascalCase should be uppercase
+      sanitized = capitalize(sanitized)
+      break
+
+    case CaseStyle.camelCase:
+      // The first character in camelCase should be lowercase
+      sanitized = uncapitalize(sanitized)
+      break
+
+    case CaseStyle.ALL_CAPS:
```
Alan-Cha: I slightly changed the behavior here. There used to be an if statement that would remove the first underscore if it exists (which would remove an extra underscore which I thought would be added by the reduce) but now I see that it is extraneous.

ErikWittern: All right.

PR #279, opened 2019-12-04
Added selectQueryOrMutationField option

gtamas:
selectQueryOrMutationField: override the GQL type (mutation  or query) of any operation.
PR contains the logic, types etc.

Tests and docs are yet to be added.


Alan-Cha:
@gtamas Awesome work!!! Thank you so much! We can iterate off of this!

Alan-Cha:
@gtamas I took a look at your PR and made some changes. I was going to push directly to your branch but I didn't have write access so I wasn't sure what to do. I ended up making my own branch and creating a PR to yours. 

For reference: https://github.com/icellmobilsoft/openapi-to-graphql/pull/1

Alan-Cha:
I'm pretty happy with the current state but I would like @ErikWittern to review before finally pulling and releasing a new version.

Alan-Cha:
Besides that, @gtamas you also need to sign off your commit. 

```
git commit --amend --signoff
```

Once you do this, you cannot accept my PR because they'll be out of sync. We can figure this out later.

gtamas:
@Alan-Cha 

Thanks for getting back to me! 
I can see you've made some minor changes and few major ones as well :) . It's cool with me, 
I'm just a bit confused now about whether this PR was good or bad in the first place :). 

I mean, for example I can see that among other things, you also changed the logic and relocated it to preprocessor. Of course this is fine, but I added it to index.ts because I tried to follow your instructions and I thought I was supposed to change the part around that if statement.
Now you basically had to rewrite that whole part.. well almost. 

I believe I could have done this myself, but I didn't know it should go to the preprocessor instead and therefore that's what I'm supposed to look at.
I could have also added the test as well..

Well anyway, I'm sure it's better this way, you know the code much better than me :). I'm just a bit confused about this now.

So now what's next? I guess we need to wait for ErikWittern to approve and then I need to sign off the commit, right?

Alan-Cha:
@gtamas The PR was very helpful for me so I think it was a good idea! I'm sorry if you felt that your work has been overwritten but that's not the way I see it. I think you introduced good ideas like in types, assigning the `selectQueryOrMutationField` option to a interface type to maximize reuse. I think it's something we need to do for the other options as well. 

And I should also apologize, once I reviewed your PR, I realized that it made more sense just to change `isMutation` in `preprocessing.ts` so it's my fault for not giving you better instruction. 

Yes, I would like to wait for @ErikWittern to review the PR and pull it in before a make a new release. 

Besides that, can you sign off? You should do it before we pull it in.

Alan-Cha:
@gtamas On second thought, if you would really like this feature right now, then I can pull it in without @ErikWittern's approval. Normally, we like to peer review PRs but I believe @ErikWittern may be out for a few days. Please let me know and I can just finish the job myself.

gtamas:
@Alan-Cha 

OK thanks for the explanation! I'm totally OK with this, it's just I was a bit confused about some aspects of the PR, but I understand it now. 

Yes it would be great if you could pull it in, thank you!

I will sign off.

Alan-Cha:
@gtamas Ok, I can try to merge it in then. Unfortunately, it seems like your Git history is a bit messed up. Is it possible to merge your commits into a single one?

I can give it a go. I will just have to submit a different PR.

Alan-Cha:
See #282

Issue #278, opened 2019-12-04
header parameters incorrectly added as operation input parameters


mikestaub:
If a parameter is expected to be passed in the request header and not the body, it should not be added as an input variable. 

Here is an example of a broken schema:

```
swagger: "2.0"
info:
  version: "1.0.0"
  title: Test API

host: localhost:1339
basePath: /

schemes:
  - http
  - https

consumes:
  - application/json
produces:
  - application/json

x-servers:
  - url: //api.test.com
    description: Default server

paths:
  /v1/files:
    post:
      parameters:
        - $ref: "#/parameters/headerparam"
      summary: Create a file
      operationId: postFile
      description: Creates a new file
      responses:
        "201":
          description: Success
          schema:
            $ref: "#/definitions/File"

parameters:
  headerparam:
    name: headerparam
    type: string
    in: header
    required: true
    pattern: "[a-zA-Z0-9-_]+"
    minLength: 1
    maxLength: 128

definitions:
  Type:
    enum:
      - TEST_ENUM
    type: string

  File:
    type: object
    properties:
      cadType:
        $ref: "#/definitions/Type"

```

Alan-Cha:
@mikestaub Once again, thank you so much for filling this issue and even providing a sample OAS to test with! I'm actually a little bit surprised that we've encountered so many issues with OtG but I'm glad we are able to debug them now with your help. I will take a look at this later today and I'm happy to report I've made progress on some of the other issues you've filed. Currently waiting for a coworker to accept my PR.

mikestaub:
Thanks for being so responsive to my issues, I really appreciate it. This library is providing a lot of value, when I get some extra cycles I will gladly contribute.

Alan-Cha:
@mikestaub Thanks for the PR! Actually, I would like to ask for some clarification. We do intend to add headers as arguments and they should be passed as a request header to the API. 

I tried to simulate a GraphQL query using your example and these were the headers that were sent. 

```
mutation {
  postFile(headerparam: "hello") {
    cadType
  }
}
```

```
{
  "headerparam":"hello",
  "content-type":"application/json",
  "accept":"application/json"
}
```

***

[...]

PR #280, opened 2019-12-04
Mikestaub/bugfix parameters

mikestaub:
https://github.com/IBM/openapi-to-graphql/issues/278

This resolves the issue, but it makes other tests fail that depend on the current functionality.

If we merge this, we have to ship a new major version as it is a breaking change.


Alan-Cha:
Closing as discussed in the related issue (#278).

We intend to produce arguments for headers but we also provide ways to avoid doing so. Will need to improve documentation, however.

PR #282, opened 2019-12-09
Added selectQueryOrMutationField option

Alan-Cha:
In correspondence with @gtamas.


packages/openapi-to-graphql/lib/types/options.d.ts
```diff
@@ -58,6 +65,17 @@ export declare type Options = {
...
     idFormats?: string[];
+    /**
+     * Allows to define the root operation type (Query or Mutation type) of any
+     * OAS operation explicitly.
+     *
+     * OtG will by default make all GET operations Query fields and everything
+     * else into Mutation fields.
+     *
+     * The field is identifed first by the title of the OAS, then the path of the
+     * operation.
```
Alan-Cha: Add method

packages/openapi-to-graphql/lib/types/options.d.ts
```diff
@@ -17,6 +17,13 @@ export declare type Report = {
     numQueriesCreated: number;
     numMutationsCreated: number;
 };
+export interface selectQueryOrMutationFieldType {
```
Alan-Cha: Perhaps, just stick with type for now

packages/openapi-to-graphql/src/preprocessor.ts
```diff
@@ -174,7 +175,33 @@ export function preprocessOas(
...
+        // Option selectQueryOrMutationField can override isMutation
+        if (
+          typeof options.selectQueryOrMutationField === 'object' &&
+          typeof options.selectQueryOrMutationField[oas.info.title] ===
+            'object' &&
+          typeof options.selectQueryOrMutationField[oas.info.title][path] ===
+            'object' &&
+          typeof options.selectQueryOrMutationField[oas.info.title][path][
+            method
+          ] === 'number'
```
Alan-Cha: Add a comment that it is a number because of TS

PR #283, opened 2019-12-09
Multiple small fixes

Alan-Cha:


packages/openapi-to-graphql/src/schema_builder.ts
```diff
@@ -870,20 +870,48 @@ export function getArgs({
...
+          // Check header option
+          if (
+            typeof data.options.headers === 'object' &&
+            parameter.name in data.options.headers
+          ) {
+            continue
+          }
+
+          // Check requestOptions option
+          if (
```
Alan-Cha: Now it checks the `requestOptions` as well

packages/openapi-to-graphql/src/schema_builder.ts
```diff
@@ -870,20 +870,48 @@ export function getArgs({
...
-        parameter.name in data.options.headers
-      ) {
-        continue
-      }
-      if (
-        typeof data.options.qs === 'object' &&
-        parameter.name in data.options.qs
-      ) {
-        continue
+      switch (parameter.in) {
```
Alan-Cha: Before, we were checking both the `qs` and the `headers` options for fulfillment. This is bad. They should have a limited scope.

packages/openapi-to-graphql/test/example_api.test.ts
```diff
@@ -1830,3 +1830,171 @@ test('Option selectQueryOrMutationField', () => {
 
   return Promise.all([promise, promise2])
 })
+
+test('Header arguments are not created when they are provided through headers option', () => {
```
Alan-Cha: Added multiple clear tests.

ErikWittern:
@Alan-Cha Nice work, looks good to me!

PR #286, opened 2020-01-03
Add oneOf and anyOf support

Alan-Cha:


Alan-Cha:
Closed in favor of #287

PR #287, opened 2020-01-04
Add anyOf and oneOf support

Alan-Cha:

PR #288, opened 2020-01-05
Fix typo in documentation for example server

cliffpyles:
The casing was incorrect for `createGraphQlSchema` in the example server implementation.


Alan-Cha:
@cliffpyles Thank you so much! Sharp eyes! Or maybe you got errors when you tried to run the code... how embarrassing! 😅 

I think we actually want to change it to `createGraphQLSchema`, with the capitalized `QL`. I can take over the changes here.

Alan-Cha:
This has been addressed in #287. Thank you again!

PR #293, opened 2020-01-23
Add requestOptions when generating linkResolvers

### LINKED ISSUE #292: LinkResolver doesn't receive requestOptions
`requestOptions` works fine with resolvers except for linkResolvers. Is there any workaround for that?

sotaokuhama:
It seems to be failing to receive `requestOptions` when generating resolvers for `links`.
Resolves #292


Alan-Cha:
Refer to #293

PR #294, opened 2020-01-23
Links and requestOptions

Alan-Cha:
Related to #292 and #293


ErikWittern:
@sotaokuhama Thanks for finding this, and creating the original PR. @Alan-Cha, thanks for taking this on and adding tests. Great to see request options now work with links!

PR #296, opened 2020-01-24
Infer names from path

Alan-Cha:


ErikWittern:
@Alan-Cha Thanks for finalizing this PR!

Issue #80, opened 2018-11-26
How to implement graphql subscriptions using oasgraph?


ryankauk:
Really interested in utilizing graphql with loopback 4 and am wondering if it's possible to implement subscriptions?

ErikWittern:
From an OASGraph point-of-view, the support of subscriptions depends on support for expressing streams / pub-sub like mechanisms in the OAS. The notion of [callbacks](https://swagger.io/docs/specification/callbacks/) and related [callback objects](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#callback-object) may be the right thing to look at, but still the debate around this seems [not to have concluded yet](https://github.com/OAI/OpenAPI-Specification/issues/55).

Apart from that, this question seems to be specifically about LoopBack 4 also. Maybe @raymondfeng or @dhmlau know more about subscription / stream / pub-sub support in LoopBack 4?

dhmlau:
@ryankauk, we're looking into supporting messaging/eventing style APIs.  Please see our story in https://github.com/strongloop/loopback-next/issues/1884.  @raymondfeng has a PoC: https://github.com/raymondfeng/loopback4-example-websocket.

dhmlau:
Actually this one might fit more what you're looking for: https://github.com/strongloop/loopback4-example-kafka. (cross posting from https://github.com/strongloop/loopback-next/issues/1925)

ErikWittern:
Related to #101

mitar:
I am looking into using GraphQL for Kubernetes API, which has a way of watching resources. I think that could be directly translated to GraphQL subscriptions. But it seems this package is currently not really supporting it. Maybe also because how Kubernetes describes [its API](https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json) looks like it simple provides `watch` parameter, and mark that the response can be `application/json;stream=watch`. So it seems this package should somehow now that if one makes a `subscribe` GraphQL, it should set `watch` parameter to true and then interpret `application/json;stream=watch` correctly?

mitar:
After more looking into this, I think that yes, OpenAPI 3.0 callbacks should be converted to GraphQL subscriptions.

[...]

PR #297, opened 2020-01-24
Parse callbacks from Open API and create subscriptions

getlarge:
After reading #80, here is a proposal on a way to handle callbacks and create corresponding subscription fields.

There is only one test for now, i would like your feedback to know if you agree with the choices made, before expanding test phases.

Signed-off-by: Edouard Maleix ed@getlarge.eu


packages/openapi-to-graphql/src/types/options.ts
```diff
@@ -132,7 +145,9 @@ export type Options = {
    * non-standard authentication requirements.
    */
   customResolvers?: {
-    [title: string]: { [path: string]: { [method: string]: ResolveFunction } }
+    [title: string]: {
+      [path: string]: { [method: string]: ResolveFunction | ResolveObject }
+    }
   }
```
getlarge: As the Subscription field wait for a required `subscribe` field and optional `resolve`, the `customResolvers` had to be adapted.

packages/openapi-to-graphql/src/types/options.ts
```diff
@@ -110,14 +116,21 @@ export type Options = {
    */
   requestOptions?: NodeRequest.OptionsWithUrl
 
+  /**
+   * Allows to override or add options to the PubSub connect object used to make
+   * publish/subscribe to the API backend.
+   * e.g. Setup the web proxy to use.
+   */
+  connectOptions?: ConnectOptions
```
getlarge: Create an option object that would follow publish / subcribe semantics instead of response / request

packages/openapi-to-graphql/src/types/operation.ts
```diff
@@ -173,6 +181,11 @@ export type Operation = {
    */
   isMutation: boolean
 
+  /**
+   * Whether this operation is a subscription.
+   */
+  isSubscription: boolean
+
```
getlarge: A flag to detect Subscription operation, set to true when the operation goes through `processOperationCallbacks` function.

packages/openapi-to-graphql/src/types/oas3.ts
```diff
@@ -173,6 +174,14 @@ type PathsObject = {
...
 }
 
+export type CallbackObject = {
+  [key: string]: PathItemObject
+}
+
+export type CallbacksObject = {
+  [key: string]: CallbackObject | ReferenceObject
+}
+
```
getlarge: Defining CallbackObject following [OpenAPI specifications](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#operationObject)

packages/openapi-to-graphql/src/types/graphql.ts
```diff
@@ -46,9 +46,27 @@ export type ResolveFunction = (
...
+  [key: string]: any
+}
+
+export type SubscriptionIterator = (
+  root: object,
+  args: object,
+  ctx: SubscriptionContext,
+  info?: object
+) => AsyncIterable<string | string[]>
+
```
getlarge: Describing the subscribe method in a Subscription field, as described by graphQL.
There seems to be conflict on either using AsyncIterator or AsyncIterable [here](https://github.com/graphql/graphql-js/issues/1442) or [there](https://www.howtographql.com/graphql-js/7-subscriptions/) or [even here](https://github.com/davidyaha/graphql-redis-subscriptions/pull/114)

packages/openapi-to-graphql/src/schema_builder.ts
```diff
@@ -17,7 +17,7 @@ import {
   ReferenceObject,
   LinkObject
 } from './types/oas3'
-import { Args, GraphQLType } from './types/graphql'
+import { Args, GraphQLType, ResolveFunction } from './types/graphql'
```
getlarge: Useless, i probably forgot to remove it after some test.

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -517,6 +781,9 @@ export function getResolver({
                   saneData = arraySaneData
                 }
 
+                // if (operation.callbacks)
```
getlarge: I was wondering if callbacks could be triggered in Mutation resolver.
Maybe through an option ?
Still it's probably safer to let them handled by user's backend ( and clearly easier ).

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -40,13 +49,272 @@ type GetResolverParams = {
...
+        /**
+         * Resolving link params may change the usedParams, but these changes
+         * should not be present in the parent _openAPIToGraphQL, therefore copy
+         * the object
+         */
+        resolveData = JSON.parse(
+          JSON.stringify(root['_openAPIToGraphQL'].data[parentIdentifier])
+        )
+      }
+    }
```
getlarge: Not sure if the above should be kept ?

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -40,13 +49,272 @@ type GetResolverParams = {
...
+            data.saneMap
+          )
+          resolveData.usedPayload = rawPayload
+        } else {
+          const rawPayload = JSON.parse(args[sanePayloadName])
+          resolveData.usedPayload = rawPayload
+        }
+      }
+    }
+
```
getlarge: Resolve request body of the callback to evaluate the path ( runtime expression ) that will be used the topic of the subscription.

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -40,13 +49,272 @@ type GetResolverParams = {
...
+        )
+      })
+      args[paramNameWithoutLocation] = value
+    }
+    const topic = args[paramNameWithoutLocation] || 'test'
+    pubsubLog(`Subscring to : ${topic}`)
+    return ctx.pubsub
+      ? ctx.pubsub.asyncIterator(topic)
+      : pubsub.asyncIterator(topic)
+  }
```
getlarge: If pubsub is found in context provided by the user, it can replace the one declared in this file.
Which is recommended to do [here](https://github.com/apollographql/graphql-subscriptions#getting-started-with-your-first-subscription)

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -40,13 +49,272 @@ type GetResolverParams = {
...
+    )
+    const customResolver = customResolvers[title][path][method] as ResolveObject
+    if (typeof customResolver.resolve === 'function') {
+      return customResolver.resolve
+    }
+  }
+
+  return (payload: any, args, context, info = {}) => {
+    // Validate and format based on operation.responseDefinition  ...
+    const typeOfResponse = operation.responseDefinition.targetGraphQLType
```
getlarge: This function requires some improvement, but probably some discussion are needed as the formatting doesn't follow REST rules.

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -40,13 +49,272 @@ type GetResolverParams = {
...
+  ) {
+    translationLog(
+      `Use custom subscribe resolver for ${operation.operationString}`
+    )
+    const customResolver = customResolvers[title][path][method] as ResolveObject
+    if (typeof customResolver.subscribe === 'function') {
+      return customResolver.subscribe
+    }
+  }
+
```
getlarge: Adapted customeResolvers conditions

packages/openapi-to-graphql/src/preprocessor.ts
```diff
@@ -155,6 +355,17 @@ export function preprocessOas(
...
+        const callbacks = Oas3Tools.getEndpointCallbacks(
+          path,
+          method,
+          oas,
+          data
+        )
+
+        // should every callback items be registered as operations ?
+        processOperationCallbacks(callbacks, oas, data, options)
+
```
getlarge: Retrieve CallbacksObject from the operationObject and resolve $refs when needed.

packages/openapi-to-graphql/src/preprocessor.ts
```diff
@@ -179,6 +390,13 @@ export function preprocessOas(
         const inViewer =
           securityRequirements.length > 0 && data.options.viewer !== false
 
+        /**
+         * Whether the operation should be added as a Subscription field.
+         * By default, all operations contained in callbacks should be Subscription ?
+         * Unless it is more logical to consider the callbacks parent to be the Subscription ?
+         */
+        let isSubscription = false
+
```
getlarge: Still up to discussion, should the subscription be the operation and the callbacks would be several channels to listen to ?
I went on describing each callback as a distinct subscription, so the parent operation is never one.

packages/openapi-to-graphql/src/preprocessor.ts
```diff
@@ -21,6 +27,196 @@ import { GraphQLOperationType } from './types/graphql'
...
+            mitigationAddendum: `The operation from the OAS '${operation.oas.info.title}' will be ignored`,
+            data,
+            log: preprocessingLog
+          })
+        } else {
+          data.operations[operationId] = operation
+        }
+
+        operations.push(operation)
+      }
```
getlarge: The operations array is probably useless, but i was not sure if we might reuse it as reference for each callback later.

packages/openapi-to-graphql/src/preprocessor.ts
```diff
@@ -21,6 +27,196 @@ import { GraphQLOperationType } from './types/graphql'
...
 const preprocessingLog = debug('preprocessing')
 
+function processOperationCallbacks(
+  callbacksObject: { [key: string]: CallbackObject },
+  oas: Oas3,
+  data: PreprocessingData,
+  options: InternalOptions
+): Operation[] {
+  const operations: Operation[] = []
+  for (let callbackName in callbacksObject) {
```
getlarge: Retrieving Operations for each PathItemObject found in CallbackObject ...

packages/openapi-to-graphql/src/oas_3_tools.ts
```diff
@@ -985,7 +1103,8 @@ export function getSecurityRequirements(
   path: string,
   method: string,
   securitySchemes: { [key: string]: ProcessedSecurityScheme },
-  oas: Oas3
+  oas: Oas3,
+  callback?: CallbackObject
```
getlarge: Added an optionnal CallbackObject to be able to reuse frunctions provided by oas_3_tools

packages/openapi-to-graphql/src/oas_3_tools.ts
```diff
@@ -166,6 +181,35 @@ export function countOperationsMutation(oas: Oas3): number {
...
+    for (let method in oas.paths[path]) {
+      if (
+        isOperation(method) &&
+        method.toLowerCase() !== 'get' &&
+        oas.paths[path][method].callbacks
+      ) {
+        for (let cbName in oas.paths[path][method].callbacks) {
+          for (let cbPath in oas.paths[path][method].callbacks[cbName]) {
+            numOps++
+            // resolve $ref ??
```
getlarge: Still unsure if we should count deeper, each path contained in a callback as distinct callback

packages/openapi-to-graphql/src/index.ts
```diff
@@ -491,11 +586,45 @@ function getFieldForOperation(
      */
     requestPayloadDef: operation.payloadDefinition,
     parameters: operation.parameters,
-
     operation,
     data
   })
 
+  if (operation.isSubscription) {
+    const responseSchemaName = operation.responseDefinition
```
getlarge: If operationType is Subscription, then create a specific field with subscribe and optionally resolve

Alan-Cha:
@getlarge Wow!!! This is an incredible amount of work! We will need some time to review this but this has been an amazing surprise!

getlarge:
I'm glad that it makes you react this way.
Indeed his feature was quite intensive and even if i wished to provide you smaller chunks, it made less sense than proposing something that handles the entire process and that can be tested.
Don't hesitate to ask for clarifications, or make some suggestions.

packages/openapi-to-graphql/src/index.ts
```diff
@@ -364,12 +383,70 @@ async function translateOpenAPIToGraphQL(
...
         }
       }
+    } else {
+      // handle subscriptions from operation.callbacks
+      // 1) cbName would be the subscription field name
+      // each paths contained in operation.callbacks[cbName]
+      // would be a channel to subscribe on the resolver
+      // but if callback object contains several operations
+      // how to be sure that the returned Graphql type would be the same ?
+      // By "forcing' a common response schema for every operation within the CB ?
```
Alan-Cha: We encountered a similar problem when creating `Query` and `Mutation` fields, which may have multiple successful response objects based on the HTTP code. In that case, we simply picked the first successful response object. Perhaps there are ways to improve this in the future. 

The other solutions would be try to create a union (as we do when resolving `oneOf`) or creating an object type with the superset of all fields (assuming that all response are objects as we do when resolving `anyOf`) or defaulting onto a JSON type. 

For now, I think we should just pick an arbitrary one and throw a warning.

packages/openapi-to-graphql/src/index.ts
```diff
@@ -491,11 +586,45 @@ function getFieldForOperation(
...
     requestPayloadDef: operation.payloadDefinition,
     parameters: operation.parameters,
-
     operation,
     data
   })
 
+  if (operation.isSubscription) {
+    const responseSchemaName = operation.responseDefinition
+      ? operation.responseDefinition.graphQLInputObjectTypeName
```
Alan-Cha: I believe this should use `graphQLObjectTypeName` instead of `graphQLInputObjectTypeName`.

packages/openapi-to-graphql/src/oas_3_tools.ts
```diff
@@ -130,6 +132,19 @@ export function countOperations(oas: Oas3): number {
     for (let method in oas.paths[path]) {
       if (isOperation(method)) {
         numOps++
+        if (oas.paths[path][method].callbacks) {
+          for (let cbName in oas.paths[path][method].callbacks) {
+            for (let cbPath in oas.paths[path][method].callbacks[cbName]) {
+              numOps++
+              // resolve $ref ??
+              // for (let cbMethod in oas.paths[path][method].callbacks[cbName][cbPath]) {
```
Alan-Cha: I think incrementing `numOp` on methods makes sense.

packages/openapi-to-graphql/src/oas_3_tools.ts
Alan-Cha: I'm not really sure either. It depends on the use cases. Do we expect people to have multiple operation objects on the same callback object? I'm guessing no... If that is the case, then we not count deeper. 

We should however ensure that an operation object exists. Let's just assume that there can only be one and throw a warning if multiple ones are found.

packages/openapi-to-graphql/src/oas_3_tools.ts
```diff
@@ -906,15 +973,65 @@ export function getParameters(
 }
 
 /**
- * Returns an array of server objects for the opeartion at the given path and
+ * Returns an hash containing the callbacks defined in the given endpoint.
```
Alan-Cha: Small type: "Returns a hash..."

packages/openapi-to-graphql/src/oas_3_tools.ts
```diff
@@ -906,15 +973,65 @@ export function getParameters(
...
+ */
+export function getEndpointCallbacks(
+  path: string,
+  method: string,
+  oas: Oas3,
+  data: PreprocessingData
+): { [key: string]: CallbackObject } {
+  const callbacks = {}
+  const endpoint: OperationObject = oas.paths[path][method]
```
Alan-Cha: Just a small nitpick, but I would call this an "operation" instead of "endpoint"

packages/openapi-to-graphql/src/oas_3_tools.ts
```diff
@@ -906,15 +973,65 @@ export function getParameters(
...
+
+  if (typeof endpoint.callbacks === 'object') {
+    let callbacksObject: CallbacksObject = endpoint.callbacks
+    for (let callbackName in callbacksObject) {
+      if (typeof callbacksObject[callbackName] === 'object') {
+        let callbackObject: CallbackObject | ReferenceObject =
+          callbacksObject[callbackName]
+        // Make sure we have CallbackObject:
+        if (typeof (callbackObject as ReferenceObject).$ref === 'string') {
+          callbackObject = resolveRef(
```
Alan-Cha: Do we need to check for references here? If I recall correctly, the only time we check for references is dealing with schemas that may be contained in the components object ([related link](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#componentsObject)). 

Although, now that I've done some digging, I see that there are many other references we do not check for in general. Might need to file an issue about this.

packages/openapi-to-graphql/src/oas_3_tools.ts
```diff
@@ -906,15 +973,65 @@ export function getParameters(
...
+          let pathItem: PathItemObject = callbackObject[expression]
+          // console.log("CHECK CALLBACK ITEM", expression, pathItem)
+          if (typeof (pathItem as ReferenceObject).$ref === 'string') {
+            pathItem = resolveRef(callbackObject[callbackName]['$ref'], oas)
+          } else {
+            pathItem = (pathItem as any) as PathItemObject
+          }
+
+          const callback: CallbackObject = { [expression]: pathItem }
+          callbacks[callbackName] = { ...callbacks[callbackName], ...callback }
```
Alan-Cha: As mentioned earlier, I think callbacks should have a unique key/name based on the path and the cbName, not just the cbName. I'd like to get the opinion of @ErikWittern but if we do go with this route, we should at least check if we are overwriting a callback and throw a warning if it is the case.

packages/openapi-to-graphql/src/preprocessor.ts
```diff
@@ -155,6 +355,17 @@ export function preprocessOas(
         // Links
         const links = Oas3Tools.getEndpointLinks(path, method, oas, data)
 
+        // Callbacks containing [key: string]:PathItemObject
+        const callbacks = Oas3Tools.getEndpointCallbacks(
```
Alan-Cha: I'm wondering if `processOperationCallbacks` should return its own operation array (instead of appending it to `data.operations`) and `callbacks` here should be set to that array instead of being set to `getEndpointCallbacks`, which essentially just returns the raw callback objects, if I understand correctly?

packages/openapi-to-graphql/src/preprocessor.ts
Alan-Cha: Hmmm... I don't think we should push the `operation` to the `data.operations` or else we will add the operation as a `Query` or `Mutation` field.

packages/openapi-to-graphql/src/preprocessor.ts
Alan-Cha: I can see this as another possibility. It really depends on how people will implement their callbacks. If the callback names are unique and provide enough context by themselves, then we can create `Subscription` fields using the raw callback names. If the names themselves do not provide enough context, then we should append the path somehow. If we expect operations to contain multiple callbacks, then we can consider creating a nested structure. 

If we had more data and a better idea of conventions and best practices, it would be easier to know what is the best way to move forward.

packages/openapi-to-graphql/src/resolver_builder.ts
Alan-Cha: This is only necessary for resolving links... I'm not entirely sure what that means in respect to callbacks.

getlarge: Right, now that i know its purpose i will remove it.

packages/openapi-to-graphql/src/preprocessor.ts
getlarge: Each operation contained in callbacks is counted as `Subscription` because these operations are filtered later with `isSubscription`.
That's why i thought it might be useless to create this new array, unless we process them differently, but it didn't seem necessary.

packages/openapi-to-graphql/src/preprocessor.ts
getlarge: Even if a `CallbackObject` as uniqueName it might still contains several paths expression, [example](https://swagger.io/docs/specification/callbacks/).
Each `PathItemObject` may have it's own parameters, requestBody and responses, so if we wish to stick to OpenAPI, the simplest case would be one `Subscription` per `PathItemObject`.
But if the resolvers are able to transform the path (expression) to create severals `Channels` to subscribe to and then to resolve the payload, it would be one `Subscription` per `CallbackObject`.

packages/openapi-to-graphql/src/oas_3_tools.ts
getlarge: It makes echo to the comment above on `preprocessor.ts`
Maybe counting one operation would be easier to start with and we could slowly iterate to count and handle properly all of them later.

packages/openapi-to-graphql/src/index.ts
getlarge: Yes it would make sense, probably a copy/paste mistake.

packages/openapi-to-graphql/src/oas_3_tools.ts
getlarge: Right

packages/openapi-to-graphql/src/oas_3_tools.ts
getlarge: Yes probably, i reused the same terminology of above functions.

Alan-Cha:
@getlarge Actually, I still haven't fully reviewed the PR. Just wanted to let you know! Sorry, I've been bogged down by other work but so far, this looks very promising! Again, this is an incredible effort! I'm sorry I haven't been able to follow up as quickly as I'd like.

packages/openapi-to-graphql/src/oas_3_tools.ts
getlarge: It allows to pseudo validate `CallbacksObject` and also to recompose them since they could be $ref to describe the `CallbackObject` and/or the `PathItemObject`.
Then we can safely go deeper to parse the operation that each `PathItemObject` contains in `processOperationCallbacks`.

Maybe i got it wrong ?

packages/openapi-to-graphql/src/preprocessor.ts
getlarge: Right, here `callbacks` are still raw.
I was also hesitating, but it seems safe to process them as operation.
After all, callbacks are composed of `OperationObject`, so why not ?

getlarge:
@Alan-Cha It's ok, no worry! 
Since these updates are prone to a lot of interpretation, it might be good to take time to think of how to implement them well. So there should be no rush.
Even if i will be happy to use these features from npm ;)

getlarge:
@Alan-Cha  Hey, do you have any idea when you will have some time to work on this project again ?
And just to let you know, i recently used my fork on a loopback 4 project to benefit from the subscription resolvers and it really works well ;)

Alan-Cha:
@getlarge Hey! I am more free this week. Will try to give you more feedback later today!

Alan-Cha:
Were you able to run your tests without failures? I just tried and encountered some issues.

Alan-Cha:
It seems to be a problem with overlapping port numbers with other test suites.

getlarge:
From what i remember, all the tests were successful before i pushed.
Tomorrow i will have some time to test it again and to participate in your review.

Alan-Cha:
I addressed the issue by just changing the port number. The one other issue is that the Aedes MQTT server has some listeners that will try to log messages after the tests have finished running. I removed these listeners and I no longer have any more problems.

packages/openapi-to-graphql/src/preprocessor.ts
getlarge: It's a good possibility, then i think we would not need the `isSubscription` flag since subscriptions would be composed from data.callbacks instead of data.operations ( in `translateOpenAPIToGraphQL` ), so there would be no need to filter them.
But i think keeping this treatment in a separate function like `processOperationCallbacks` allow to keep things clear in `preprocessOas`.

packages/openapi-to-graphql/src/oas_3_tools.ts
getlarge: Since `getEndpointCallbacks` is called before `processOperationCallbacks` and that the last one needs information about the whole callback content to compose the callback operation that is saved in `PreprocessingData`, i thought it was required to resolve all the potential references.
What would be the alternative then ?

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -40,13 +49,272 @@ type GetResolverParams = {
...
+      paramNameWithoutLocation = paramName.split('.')[1]
+    }
+
+    // /**
+    //  * see if the callback path contains constants expression
+    //  *
+    //  */
+    if (value.search(/{|}/) === -1) {
+      args[paramNameWithoutLocation] = isRuntimeExpression(value)
+        ? resolveLinkParameter(paramName, value, resolveData, root, args)
```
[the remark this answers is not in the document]
getlarge: Sure point, that's probably why you got misled in your previous comment.
But since the callback path is a runtime expression, like the link parameter ...

packages/openapi-to-graphql/test/example_api5.test.ts
```diff
@@ -0,0 +1,166 @@
...
+import * as openAPIToGraphQL from '../lib/index'
+import { Options } from '../lib/types/options'
+import { startServers, stopServers } from './example_api5_server'
+import { GraphQLOperationType } from '../lib/types/graphql'
+
+const oas = require('./fixtures/example_oas5.json')
+
+const TEST_PORT = 3000
+const HTTP_PORT = 3005
+const MQTT_PORT = 1885
```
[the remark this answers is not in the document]
getlarge: Yes it was like, test number 5 ? so i append 5 to the port.
Not really deep reason ;)

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -40,13 +49,272 @@ type GetResolverParams = {
...
+          resolveData.usedPayload = rawPayload
+        }
+      }
+    }
+
+    pubsubLog(
+      `Subscription schema : ${JSON.stringify(resolveData.usedPayload)}`
+    )
+
+    if (typeof resolveData.usedParams === 'undefined') {
```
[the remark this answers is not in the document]
getlarge: Since resolveData.usedParams is used in the future `resolveRuntimeExpression`, we will probably still need it.

packages/openapi-to-graphql/package.json
```diff
@@ -40,8 +40,10 @@
   "types": "lib/index.d.ts",
   "scripts": {
     "api": "nodemon test/example_api_server.js",
+    "api-5": "nodemon test/example_api5_server.js",
```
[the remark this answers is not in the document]
getlarge: I agree, it was to make my tests faster.

packages/openapi-to-graphql/src/auth_builder.ts
```diff
@@ -107,7 +108,12 @@ export function createAndLoadViewer(
     }
 
     // Create name for the viewer
-    let viewerName = !isMutation
+    let viewerName = isSubscription
```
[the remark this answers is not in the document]
getlarge: If we bring callbacks into a separate field of `PreprocessingData` we won't need `isSubscription` anymore.

getlarge:
Sorry for that, i also met the issues with the tests, my previous computer was probably too slow to encounter it ! ( port conflict and aedes listener ).

I hope you appreciate this occasion to learn about pubsub communications.

You mentioned code duplication, what do you have in mind ?

Alan-Cha:
@getlarge I have some basic questions that I would like to ask you to help my understanding. I'm not very experienced in this area which is part of the reason why my response is very slow.  

[graphql-subscriptions](https://www.npmjs.com/package/graphql-subscriptions) works with different network transports and mentions [subscriptions-transport-ws](https://www.npmjs.com/package/subscriptions-transport-ws) as an example but are you aware if there are any other network transports that are supported? I had trouble finding any other examples. If not, are you aware of any efforts to do so? 

I can see that this PR uses the key in the [callbacks object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#callbackObject) as the topic of the pubsub connection. According to the documentation, “The key value used to identify the path item object is an expression, evaluated at runtime, that identifies a URL to use for the callback operation.” I wonder if the practice of using the key to define the topic of a pubsub connection instead of a URL is a common practice.

Alan-Cha:
On code duplication, don't worry about it. I'm currently making the changes.

getlarge:
@Alan-Cha `graphql-subscriptions` provides a PubSubEngine abstract class that can be extended, it's using an eventEmitter and as you can read at the bottom of the README.md of this project there are several implementations with different pubsub engines ( MQTT, Redis ...). 
It's important to note that hose implementations behave like clients, like you can see on the example i published `example_api5_server.js`, where i used a MQTT broker ( Aedes ) and a separated MQTT client wrapped in a PubSubEngine (`graphql-mqtt-subscriptions`).

`subscriptions-transport-ws` is a special case, as it provides a websocket server and a client in the same library.

If you want to see a rather simple example of using graphql subscriptions combined with an MQTT broker, you can take a look at this project : [entourage](https://github.com/chriskalmar/entourage/).
And another one, a bit more complex, that uses `openapi-to-graphql` and `loopback 4`  [device-manager-graphql](https://github.com/aloes-io/device-manager-graphql).
We can schedule a code review if that might help you to understand faster.

Concerning the callback object, using the path (as runtime expression) was the whole trick and it's not a common practice because it's not obvious. Originally the openAPI spec is not made for async communication, but by doing so we can resolve dynamically the topic and make it more compatible with pubsub protocols.

Alan-Cha:
@getlarge Thanks for clarification. I see that there are many other pubsub implementations that are available. This is good! We want to use something general and flexible. 

I think a code review would be very helpful! My colleague @ErikWittern also mentioned he would like to participate. In particular, there seems to be a problem with the test suite where there are asynchronous operations that have not been properly stopped when it ends. Maybe we can debug this together. In any case, what is a good time for you?

I see. I think we should consider making this feature an option instead of default behavior for people who may have callbacks in the OAS for other reasons. Otherwise, they may see weird things in their schemas. It could be called something like `turnCallbacksIntoSubscription`.

getlarge:
@Alan-Cha The test suite is not really serious and profound to be honest, i just wanted to demonstrate the principles with a network configuration.
It might be more relevant to use directly the pubsub instance instead, like [this](https://github.com/apollographql/graphql-subscriptions/blob/master/src/test/tests.ts).

And i totally agree on a `turnCallbacksIntoSubscription` option or any other eloquent name.

Would you like to start this review with some projects using graphQL subscriptions ?
I'll be available tomorrow and this weekend, or any end of the day next week ( GMT +1 ).

Alan-Cha:
@getlarge I see! We can discuss it in more detail tomorrow. 

I am in GMT -4 and @ErikWittern is GMT +2, I believe. Unfortunately he is offline right now but hopefully we can hear back from him tomorrow morning. I will also try to get up early so we can plan appropriately. My schedule is generally free except one meeting at 11 AM. 

Perhaps for now, we can tentatively plan for 9 AM for me, or 1 PM for you, @getlarge, and 2 PM for @ErikWittern. Does that work for you?

We're very excited to discuss this with you! Thank you for your compliance!

ErikWittern:
@getlarge @Alan-Cha 2pm GMT+2 works for me. Looking forward to it!

getlarge:
@Alan-Cha @ErikWittern I sent you an email.

getlarge:
Finally, i left the long name `createSubscriptionsFromCallbacks` in case we find another way of generating subscription from OpenAPI schema, who knows!
Also :
- Made small changes that @Alan-Cha suggested
- Complete the doc that @ErikWittern has provided

Alan-Cha:
@getlarge I made a [PR](https://github.com/getlarge/openapi-to-graphql/pull/1) on your branch in which I reduced the amount of code duplication. To summarize, the preprocessing data now contains a new field `callbackOperations` similar to `operations` but instead only contains the `Operation`s from callbacks. In addition, I refactored `preprocessOas()` and created a new function `preprocessOperation()` that is used to process both the operation objects and callback objects. Furthermore, it no longer relies on `isMutation` and `isSubscription` but rather relies on the new `operationType` field. Lastly, I made a number of changes to documentation and function names to make them more consistent across the board. 

I believe this can be pulled in to your branch. Let me know what you think. In the meantime, I will review the changes you've made since our last talk. Thanks!

packages/openapi-to-graphql/README.md
```diff
@@ -174,6 +174,8 @@ Resolver options:
...
 
+- `createSubscriptionsFromCallbacks` (type: `boolean`, default: `false`): Allow to generate subscription fields from CallbackObjects in OpenAPI schema. Path ( runtime expression ) of the [CallbackObject](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#CallbackObject) will be interpolated as topic of publish / subscription to use with a [pubsub](https://github.com/apollographql/graphql-subscriptions) instance.
```
Alan-Cha, line 177: I would like to suggest some wording changes

```
Generates subscription fields from [callback objects](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#CallbackObject). The keys ([runtime expressions](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#runtimeExpression)) of the callback objects will be interpolated as the topic of a publish/subscription connection using [graphql-subscriptions](https://github.com/apollographql/graphql-subscriptions). Read the [doc](./docs/SUBSCRIPTIONS.md) for explanations and examples regarding its usage.
```

packages/openapi-to-graphql/docs/SUBSCRIPTIONS.md
```diff
@@ -0,0 +1,279 @@
+# Subscriptions with OpenAPI-to-GraphQL
```
Alan-Cha, line 1: Should this tutorial live in the `openapi-to-graphql` package? Or in the monorepo like with the other tutorials/docs?

packages/openapi-to-graphql/docs/SUBSCRIPTIONS.md
```diff
@@ -0,0 +1,279 @@
...
+Since version x.x.x, OpenAPI-to-GraphQL supports [GraphQL _subscription_ operations](http://spec.graphql.org/draft/#sec-Subscription). In GraphQL, using a subscription query, clients subscribe to updates on the data defined in the query. In this scneario, when data changes, the server publishes these changes to all clients that have active subscriptions for that data.
```
Alan-Cha, line 2: This will most likely be version 2.1.0.

packages/openapi-to-graphql/docs/SUBSCRIPTIONS.md
```diff
@@ -0,0 +1,279 @@
...
+
+The OpenAPI specification can define similar behavior using [callbacks](https://swagger.io/specification/#callbackObject): a callback defines a request that the server may initiate in response to receiving another request. Callbacks can thus be used to model publish/subscribe behavior. I.e., when the server receives a request to update some data, it can then itself issue callback requests (outside of the first request/response cycle) to any number of subscribed clients to inform about the new data.
+
+When enabling the `createSubscriptionsFromCallbacks` option, OpenAPI-to-GraphQL creates a subscription field for any operation in the given OpenAPI document that defines a `callback` object. In such cases, OpenAPI-to-GraphQL creates a [subscribe](http://spec.graphql.org/draft/#Subscribe()) function responsible to subscribing clients to receive results of callbacks being executed, and a special form of a [resolve](http://spec.graphql.org/draft/#ResolveFieldEventStream()) function, which pushes data updates to subscribed clients.
```
Alan-Cha, line 6: I would like to add a link to the options like the following

```
[`createSubscriptionsFromCallbacks` option](https://github.com/IBM/openapi-to-graphql/tree/master/packages/openapi-to-graphql#options)
```

In addition, I think we need to add a bit more clarification. You can define multiple callbacks on an operation, multiple path items on a callback, and multiple operations on a path item. I believe in our current implementation, we are assuming (but not checking for) that there is only operation defined per path item. Even if that is the case, the wording needs to be changed because OtG would create a subscription field for all callback objects, not just one for any operation that defines callbacks as it currently suggests.

packages/openapi-to-graphql/docs/SUBSCRIPTIONS.md
```diff
@@ -0,0 +1,279 @@
...
+```javascript
+import { EventEmitter2 } from 'eventEmitter2';
+import { PubSub } = from 'graphql-subscriptions'
+
+const eventEmitter = new EventEmitter2({
+  wildcard: true,
+  delimiter: '/'
+});
+
+// Create the PubSub instance ( here by wrapping an EventEmitter client )
```
Alan-Cha, line 25: I would like to remove the extraneous spaces inside the parenthesis 

```
// Create the PubSub instance (here by wrapping an EventEmitter client)
```

packages/openapi-to-graphql/docs/SUBSCRIPTIONS.md
```diff
@@ -0,0 +1,279 @@
...
+
+PubSub could also wrap an MQTT client connected to a broker, like in this [example API](../test/example_api5_server.js).
+
+```javascript
+import { connect } = from 'mqtt'
+import { MQTTPubSub } = from 'graphql-mqtt-subscriptions'
+
+const MQTT_PORT = 1883
+
+// Create a PubSub instance ( here by wrapping a MQTT client )
```
Alan-Cha, line 39: Same thing here regarding the extraneous spaces

```
 // Create a PubSub instance (here by wrapping a MQTT client)
```

packages/openapi-to-graphql/docs/SUBSCRIPTIONS.md
```diff
@@ -0,0 +1,279 @@
...
+
+  // ...and set up the WebSocket for handling GraphQL subscriptions
+  wsServer.listen(HTTP_PORT, () => {
+    new SubscriptionServer(
+      {
+        execute,
+        subscribe,
+        schema,
+        onConnect: (params, socket, ctx) => {
+          // adding pubsub to context
```
Alan-Cha, line 89: Changing the comment so that it matches the others...

```
// Add pubsub to context to be used by graphQL subscribe field
```

packages/openapi-to-graphql/docs/SUBSCRIPTIONS.md
```diff
@@ -0,0 +1,279 @@
...
+To create these two functions, OpenAPI-to-GraphQL relies on the popular [graphql-subscriptions](https://github.com/apollographql/graphql-subscriptions) package, which provides a unified API to support different network transports (like WebSockets, MQTT, Redis etc. - see this [list of supported transports](https://github.com/apollographql/graphql-subscriptions#pubsub-implementations)).
+
+A typical example of using OpenAPI-to-GraphQL to create a GraphQL server supporting subscriptions may look like this:
+
+### Creating PubSub instance 
+
+First, initialize a PubSub instance to spread events between your API and the GraphQL Server, in a `pubsub.js` file.
+
+```javascript
+import { EventEmitter2 } from 'eventEmitter2';
```
Alan-Cha, line 17: I don't understand why this needs to be "eventEmitter2". Is the 2 necessary?

packages/openapi-to-graphql/docs/SUBSCRIPTIONS.md
```diff
@@ -0,0 +1,279 @@
...
+    )
+  })
+}
+
+init()
+```
+
+## API server
+
+A simple example could be the following, when an HTTP client tries to create a device ( via `post('/api/devices')` route ) an event is published by the PubSub instance.
```
Alan-Cha, line 107: I would like to remove the extraneous spaces inside the parenthesis. 

```
A simple example could be the following, when an HTTP client tries to create a device (via `post('/api/devices')` route) an event is published by the PubSub instance.
```

packages/openapi-to-graphql/docs/SUBSCRIPTIONS.md
```diff
@@ -0,0 +1,279 @@
...
+        const device = req.body
+        Devices[device.name] = device
+        const packet = {
+          topic: `/api/${device.userName}/devices/${req.method.toUpperCase()}/${
+            device.name
+          }`,
+          payload: Buffer.from(JSON.stringify(device))
+        }
+
+        // use pubsub to publish the event
```
Alan-Cha, line 150: To match the other comments in this document

```
// Use pubsub to publish the event
```

packages/openapi-to-graphql/docs/SUBSCRIPTIONS.md
```diff
@@ -0,0 +1,279 @@
...
+
+startClient()
+```
+
+In this example, we rely on the [`subscriptions-transport-ws` package](https://github.com/apollographql/subscriptions-transport-ws) to create a `SubscriptionServer` that manages WebSockets connections between the GraphQL clients and our server. We also rely on the `graphqlExpress` server provided by the [`apollo-server-express` package](https://github.com/apollographql/apollo-server/tree/master/packages/apollo-server-express) to serve GraphQL from Express.js.
+
+
+Sidenote concerning callback, as you can see in the example Callback, the path (runtime expression) `/api/{$request.body#/userName}/devices/{$request.body#/method}/+` is delimited by `/` and ends with `+`, these symbols are interpreted as delimiters and wildcard when using MQTT topics.
+It needs to be adapted accordingly to the client wrapped in your PubSub instance, for eventEmitter2 you can use `*` and define your own delimiter.
+An helper might be provided in the future, to simplify this process.
```
Alan-Cha, line 265: Small typo "A helper" instead of "An helper"

packages/openapi-to-graphql/docs/SUBSCRIPTIONS.md
```diff
@@ -0,0 +1,279 @@
...
+  })
+}
+
+startClient()
+```
+
+In this example, we rely on the [`subscriptions-transport-ws` package](https://github.com/apollographql/subscriptions-transport-ws) to create a `SubscriptionServer` that manages WebSockets connections between the GraphQL clients and our server. We also rely on the `graphqlExpress` server provided by the [`apollo-server-express` package](https://github.com/apollographql/apollo-server/tree/master/packages/apollo-server-express) to serve GraphQL from Express.js.
+
+
+Sidenote concerning callback, as you can see in the example Callback, the path (runtime expression) `/api/{$request.body#/userName}/devices/{$request.body#/method}/+` is delimited by `/` and ends with `+`, these symbols are interpreted as delimiters and wildcard when using MQTT topics.
```
Alan-Cha, line 263: Some wording changes

```
Concerning callbacks, as you can see in the example
```

Alan-Cha:
@getlarge @ErikWittern I merged Edouard's PR with my PR and rebased it and submitted a new PR #311. I think we should continue the conversation there.

packages/openapi-to-graphql/docs/SUBSCRIPTIONS.md
getlarge: If there's already a tutorial folder, it would make sense to put it here. I forgot about it since i only work on the package level.

packages/openapi-to-graphql/docs/SUBSCRIPTIONS.md
getlarge: It's not necessary, but eventEmitter2 allows to use wildcard in the topic chain. Which is very useful when you compose a topic when you have variable field that cannot be determined in advanced, but which are part of the topic structure. I mentioned it in the SUBSCRIPTION doc.

packages/openapi-to-graphql/docs/SUBSCRIPTIONS.md
Alan-Cha: Okay I'll move it then!

Alan-Cha: Actually I decide to keep it the same place.

Alan-Cha:
Sorry for the delays. I had some deliverables for other work last week and I could not look into this at all. 

@getlarge Thanks for fixing the bug! I would actually recommend that we close this PR in favor of #311 as it has been properly rebased. I pulled over the changes you made and also made additional changes. Now I think we can finally get this in and make a new release! I would like @ErikWittern to do a final review, however.

Alan-Cha:
I do have one concern and I'm sorry I haven't brought it up until now. I'm not sure if using `customResolvers` on callbacks is a good idea. Currently `customResolvers` is performing double duty on both Query/Mutation resolvers and Subscription resolvers. If someone would like to introduce custom resolvers for Subscription operations, it should come through a different means. I think we should remove the `customResolvers` for Subscription resolvers. If someone asks for the feature, we can add it in a different form.

PR #303, opened 2020-02-07
Expose tags, original operation and preprocessing data

dotansimha:
Hi :) 

I'm working on a tool that uses `openapi-to-graphql` internally, and I needed a few minor changes in the source code:

* Expose `tags` from source OAS as part of the operations. This is useful because other OAS tools uses tags. 
* Expose the raw OAS operation object. 
* Allow access to the preprocessing (IR) object to make it simple to access data. 

As a side note, a few points to improvement: 
* I think this tool should leverage GraphQL `extensions` feature (https://github.com/graphql/graphql-js/pull/2097) and keep the OAS metadata as part of the created schema. It will make it simpler to get the source OAS object (instead of trying to guess it from the operation name or field names). 
* I understand that adding compiled `lib` to the repo allow developers to be able to install this tool directly from GitHub, but it also causes issues with rebases and merge changed - maybe it could be better to publish alpha version per each commit (we are doing something similar in `graphql-code-generator` repo) instead?  

Thank you for this tool, it's really helpful, and it's approach really simplifies the integration between GraphQL and OAS :)


Alan-Cha:
@dotansimha Sorry for the late reply but thank you so much for this PR! I like these changes! I think exposing the preprocessing object and the `extensions` idea would greatly boost usability. I'm not entirely sure if we should expose everything in the preprocessing object, however. I will have to take another look. 

Yeah, we aren't exactly sure how we should present the code on GitHub and in our NPM package. We're open to any advice. I will take a look at your repo! It seems like an interesting idea. 

All in all, thank you so much for contributing and your kind compliments! Hope you continue using OtG in the future!

Alan-Cha:
@dotansimha I really dropped the ball on this one. It's been over an entire year. I had a really hectic year last year that prevented me from working on OtG as consistently as I wanted to and I just forgot to follow up. It's not really an excuse but it's the best I can come up with. I'm really sorry about that. Hope you were still able to use OtG and thanks for the contribution!

PR #311, opened 2020-04-06
Parse callbacks from Open API and create subscriptions

Alan-Cha:
Related to #297


Alan-Cha:
I had to rename the new API to Example API 7 as there already exists and 5 and a 6. 

Unfortunately, I am still having some issues with the test. I get the following error from Jest when I run the test in isolation: 

```
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

Try running `npm test api7` and let me know if you get the same error. I'm not exactly sure how to resolve this issue. It also seems to be a pain to use the suggested option as I'm not sure how to configure the Jest CLI to work with TypeScript.

getlarge:
@Alan-Cha the code changes looks good ;)
I will give you more feedbacks on friday.

getlarge:
I merged your changes, and fixed the test for api7. The TCP server was not closed properly, i also changed jest config as it might be useful in the future to track those workers not closed properly.
The new `processOperation` method is really nice !

Alan-Cha:
I've create a new `README.md` file to explain the different example APIs. Hopefully this will at least clear up some of the confusion.

getlarge:
@Alan-Cha Good we're almost there ;)

For the customResolvers in Subscription, i don't see where the conflicts could happen, could you explain please ?
Since callbacks and those specific resolvers would be used together only if the option `createSubscriptionsFromCallbacks` is true, the user should be conscious that his callbacks will be dedicated to create Subscription fields ?

Alan-Cha:
@getlarge Hey! Some tests were still failing after your changes but I managed to get them working again! I thought it may have to do with inconsistent package versions across the different testing environments so I played around with the configuration and it seemed to do the trick! 

My concern with `customResolvers` is that in the current iteration, you can only define a resolver for either the operation or the callback (that is on the operation) but you cannot do both. I think I will simply create a new option called `customSubscriptionResolvers` and move things around to make it explicitly clear. Thoughts?

getlarge:
@Alan-Cha I will try the tests on your branch so that you have some additional feedback for the 'trick'.

Thanks for the explanation on the `customResolvers`, i didn't notice this problem.
Anyway it makes sense to have another object `customSubscriptionResolvers` since they won't have the same properties and both resolvers could even have their own type.

Alan-Cha:
@getlarge Don't worry about the tests! In my opinion, we should have tried to keep package versions consistent in our Travis tests from the beginning. There are no more failing tests now so I think we should move on. I also could not replicate the failures that were showing up in Travis previously. 

I added the `customSupsctionResolvers` option so I think we're pretty much set to pull and publish!

packages/openapi-to-graphql/README.md
```diff
@@ -176,9 +178,9 @@ Resolver options:
...
...us a triply nested object where the outer key is the title, followed by the path, and finally the method, which points to the [resolver function](https://graphql.org/learn/execution/#root-fields-resolvers) itself. The resolver function can use the parameters `obj`, `args`, `context`, and `info` in order to produce the proper data, as do standard [resolver functions](https://graphql.org/learn/execution/#root-fields-resolvers) in GraphQL. Use cases include the resolution of complex relationships between types, implementing performance improvements like caching, or dealing with non-standard authentication requirements. _Note: Because the arguments are provided by the GraphQL interface, they may look different from the [parameters](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#parameterObject) defined by the OAS. For example, they will have [sanitized](https://github.com/Alan-Cha/openapi-to-graphql#characteristics) names. The [request body](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#requestBodyObject) will also be contained in the arguments as an [input object type](https://graphql.org/graphql-js/mutations-and-input-types/)._
```
getlarge, line 183: It makes it clear like this ;)

packages/openapi-to-graphql/src/index.ts
```diff
@@ -519,7 +517,8 @@ async function translateOpenAPIToGraphQL(
     }, 0)
 
   /**
-   * Organize created queries / mutations / subscriptions into viewer objects.
+   * Organize authenticated Query, Mutation, and Subscriptions fields into
+   * viewer objects.
```
getlarge, line 521: I think that quite soon we will see the issue of the auth for Subscription, because for now i think it's a rather limited because it's based on HTTP security schemes that will only apply to Websocket eventually.
Usually for a subscription, the authentication takes place a connection time, from parameter set in the PubSub client instance. But then Pubsub instance like eventEmitter don't take auth parameter...
So those authViews could set the connection params, if it's possible ... As you can see it might require some thoughts and tests.

packages/openapi-to-graphql/README.md
Alan-Cha: Awesome!

packages/openapi-to-graphql/src/index.ts
Alan-Cha: Yes, I see! We should file an issue for this. I can get started on it and if you'd like to add anything, feel free to do so! But for this PR, I think we can go ahead :)

Alan-Cha:
I made some final changes generally centered around comments. Once the tests run, I will publish!

Alan-Cha:
Looks like the tests have finished running! It's finally in! Thank you so so so much again! I know it's taken a long time to get to this point but I'm happy we finally got here! This has been the biggest PR we've ever had so it is really an incredible achievement. It has been a pleasure working with you @getlarge 🥳🍾🎆

Alan-Cha:
And it's out!!!

[Release notes](https://github.com/IBM/openapi-to-graphql/releases/tag/openapi-to-graphql%402.1.0)

getlarge:
Very nice !! I already updated some of my package.json ;)
@Alan-Cha Pleasure was mine too and thanks for your improvements and your time spent on reviews.
See ya on the next issue / PR ;)

Issue #308, opened 2020-02-25
Query strings improperly escaped


pbmacg:
**Describe the bug**
I'm trying to use the deepObject & explode parameter serialization options to create a wrapper around a loopback-like API that exposes exactly their ?filter[where][x][]=y syntax. I've got just about everything working, but when the requests from GraphQL hit the REST API the query strings come out looking like: `[25/Feb/2020:16:51:14 +0000] "GET /v2/posts?filter%5Bwhere%5D%5Bchannel%5D%5Bin%5D%5B0%5D=60771 HTTP/1.1" 200 846 "-" "openapi-to-graphql"`

Which, when decoded, is almost correct (has a 0 index in the final set of square brackets, not sure if the API will like that yet, but one problem at a time). 

My problem is that the API doesn't seem to understand query strings when presented in this escaped format. So my question is this-- how can I prevent this string escaping?

**To Reproduce**
1. Use the OAS definition included at the end of this isssue.
2. Notice that GET requests escape the structure of their query params when run

**Expected behavior**
The query parameter special characters (specifically [ & ]) shouldn't be escaped.

**OAS Definition**
`openapi: 3.0.0
info:
    version: 1.0.0
    title: MyGreatAPI V2
    description: Main API

servers:
    - url: https://api.mygreapapi.com/v2

components:
    securitySchemes:
        bearerAuth:
            type: http
            scheme: bearer
            bearerFormat: JWT
    parameters:
        filter:
            in: query
            name: filter
            style: deepObject
            explode: true
            schema:
                type: object
                properties:
                    where:
                        type: object
                    fields:
                        type: object
                    offset:
                        type: integer
                        minimum: 0
                    limit:
                        type: integer
                        minimum: 0
                    order:
                        type: array
                        items:
                            type: string
                    in:
                        type: array
                        items:
                            type: string

paths:
    /posts:
        get:
            description: Get posts
            parameters:
                - $ref: '#/components/parameters/filter'
            responses:
                '200':
                    description: Your posts sir
                    content:
                        application/json:
                            schema:
                                type: object
                                properties:
                                    data:
                                        type: object
                                        properties:
                                            id:
                                                type: string
`

pbmacg:
Traced the issue back to the querystring library being used by the node Request module. Haven't found a nice solution, but unescaping the result of the stringify method defined in Request's querystring.js does the trick.

Alan-Cha:
[...]

PR #312, opened 2020-04-17
Adds query string serialization optimization

akoenig:
👋

This PR introduces the serialization of the query string, when the `explode` attribute has been set. When defined the actual serialization includes a duplication of the parameter. Otherwise arrays will be serialized in a comma-separated way.

Example with `explode` turned off:

```
http://localhost:3004/api/robots?types=Droid%2CBot
```

Example with `explode` turned on:

```
http://localhost:3004/api/robots?types=Droid&types=Bot
```

This also activates the `useQuerystring` mode in `request` so that the query string gets encoded properly (#308).

I ran into the "problem" today, when using query string parameters which are typed as `arrays` and wanted to help to fix that. So this is my PR 🙂


Alan-Cha:
@akoenig This is an awesome change!!! Thank you so much! We're in the middle of trying to get our 2.1.0 version out and I will try to get this in right after!

papatego123:
how will this work when you have a nested object ?
```
> q = {filter: { first_name: 'hello', last_name: 'world'}}
{ filter: { first_name: 'hello', last_name: 'world' } }
> querystring.stringify(q)
'filter='
```

PR #315, opened 2020-04-26
Replace request with isomorphic-unfetch

JsonKim:
request has been deprecated since Feb 2020. And I want to use openapi-to-graphql on client-side. However, I can't because request uses fs. So I make and open this pr.


Alan-Cha:
@JsonKim This is a very important change! Thank you very much! 

I'm not familiar with `request` alternatives. Could you give us some insight into why this particular library was selected?

ErikWittern:
@JsonKim @Alan-Cha I guess one starting point is to look at the list of alternative libraries that the request team itself compiled: https://github.com/request/request/issues/3143

JsonKim:
@ErikWittern Thanks! @Alan-Cha unfetch is lighter and simpler than other alternatives. And it can be used in both server-side and client-side because it supports polyfill. If you suggest another alternative, I can work on that as well.

PR #325, opened 2020-06-12
Don't set the content-type when there is no payload

lionelschiepers:
Hi,

I can't use the Mock Web Server https://github.com/stoplightio/prism because the GET operations set a content-type & content-length in the http headers. I think it's an error. a content-type & content-length must be set when there is a payload with operations like 'PUT' or 'POST'...

Web server like https://github.com/stoplightio/prism (mock of openapi) rejects the calls because it validate the request and a content-type has nothing to do with GET operations.

Exemple of a call made by openapi-to-graphql to my openapi web server:

GET http://127.0.0.1:4010/RESTMethod HTTP/1.1
content-type: application/json (bad)
accept: application/json
host: 127.0.0.1:4010
content-length: 0 (bad)
Connection: close


lionelschiepers:
Hi, I'm not a specialist of node.js/javascript.

I tried to find the reason why payloadRequired was not correctly initialized during the unit test but didn't find the reason. That's why I changed the logic by checking the method instead of payloadRequired. Do you have any advise?

Lionel.

Alan-Cha:
@lionelschiepers I apologize for the late reply! A lot of things have been going on at my end. 

We set the `content-type` [here](https://github.com/IBM/openapi-to-graphql/blob/1e845bfe1266e8097cb79c3b00a664793b7d46ac/packages/openapi-to-graphql/src/resolver_builder.ts#L462). This will be probably a super simple fix, we just need to check for the `operation.method` to see if it is a `GET` request or not. I'm not sure where we set the `content-length`, however. I assume that's set by the `request.js` library by default when we do not include a payload. 

Let me give it a shot...

Alan-Cha:
Ah sorry, I got a bit ahead of myself. It seems that you have already made the changes. 

This is the way I would have done as well. I do not think we should make things more complicated by checking if the payload is required.

Do these changes address your issue? Do you have any other problems with OtG?

lionelschiepers:
Hi,

Yes. This pullrequest solves my problem. If every body agree on the change I would be the most happiest man in the world if the pullrequest is merged :-)

Thanks,
Lionel.

Alan-Cha:
@lionelschiepers I can try to get this checked in later today. I think we need a test case which I will look into.

PR #332, opened 2020-06-23
Suport arbitrary JSON pointer references

mtth:
First off, thank you for writing this tool. It has been very useful to my team.

We ran into an issue recently when importing OpenAPI specs which contain references (for example, [PagerDuty's](https://raw.githubusercontent.com/PagerDuty/api-schema/master/reference/REST/openapiv3.json)). With this change, we were able to import them as expected. The root cause seems to be that references in API specs can contain characters which require escaping as described in [RFC 6901](https://tools.ietf.org/html/rfc6901) (forward slashes are particularly common in API specs since they are used as path item key). This PR leverages `json-ptr` to handle this and support arbitrary references.

Signed-off-by: Matthieu Monsch <matthieu.monsch@gmail.com>

### DIFF
--- a/packages/openapi-to-graphql/package-lock.json
+++ b/packages/openapi-to-graphql/package-lock.json
@@ -6511,6 +6511,11 @@
       "integrity": "sha512-mrqyZKfX5EhL7hvqcV6WG1yYjnjeuYDzDhhcAAUrq8Po85NBQBJP+ZDUT75qZQ98IkUoBqdkExkukOU7Ts2wrw==",
       "dev": true
     },
+    "json-ptr": {
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/json-ptr/-/json-ptr-1.2.0.tgz",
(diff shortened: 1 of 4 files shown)


tovbinm:
@Alan-Cha please have a look

ErikWittern:
@mtth Thank you for the PR, it is important that we get these things right.

I think that the specific issue you point out could also have been fixed by an additional call to [`unescape`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/unescape) at the beginning of `resolveRef`. For example, `unescape('#/paths/~1users~1%7bid%7d/description')` (from your test case) yields `'#/paths/~1users~1{id}/description'`. As such, there is a tradeoff here between a small fix for the specific issue vs. introducing a new runtime dependency. I do like that `json-ptr` has itself no dependencies, though, and is relatively small with ~90kb (as long as we are not talking browser). And I guess it will resolve other (edge-) cases that our `resolveRef` (obviously) currently misses. So LGTM!

Alan-Cha:
@mtth Thank you so much for this contribution!!! 🥳

tovbinm:
Thank you for a quick review!

mtth:
Happy to contribute - thank you for the quick review.

PR #341, opened 2020-07-31
Allows string values for `genericPayloadArgName` and 'input' for `simpleNames`.

jun-sheaf:
The changes made are quite reasonable. Hopefully you agree.

1. I have allowed the `genericPayloadArgName` value to be any non-empty string rather than just a `boolean`. Old syntax works of course, but now writing a string (e.g. `input` as written in GQL's manual) will change the payload argument name to precisely the entered string.

2. In general, `simpleNames` is better off, but when playing around with `input` types, there is some strange (but understood) behavior; if you have a model named `...Input` and you only use that for request bodies, the actual output is `...Input2Input` when you probably want just `...Input`. This happens as a result of the original model not being used as a response payload. To obfuscate this, I have implemented an `input` string option for `simpleNames` to prevent processing the **first** input model the library runs across. Subsequent input models of the same name will be processed normally.


Alan-Cha:
@jun-sheaf Hey! Thank you so much for submitting a PR! Sorry for the delay, I've been very busy with other work but I finally got around to take a look this.

I think the `genericPayloadArgName` change is reasonable. This is actually something we've considered when we first created the option but the reason why we didn't go forward with it was that we didn't think creating an option that can take both a `boolean` or a `string` input was very clean. We decided to hold off on the feature and stick to a hard coded string until someone requested the feature. Seeing that you made this a PR proves that this is a necessary feature and I cannot think of any better way to express it so I am fine with this change.

Regarding the second point, I have a some reservations. First, I don't think that this feature should be attached to the `simpleNames` option. The `simpleNames` option is only intended to control the casing of names. Instead, I think this should be a separate option.

packages/openapi-to-graphql/src/preprocessor.ts
```diff
@@ -1065,7 +1073,10 @@ function getSchemaName(
     while (usedNames.includes(`${schemaName}${appendix}`)) {
       appendix++
     }
-    schemaName = `${schemaName}${appendix}`
+
+    if (!input || appendix !== 2) {
```
Alan-Cha, line 1077: Secondly, I definitely do not like this change. The purpose of this while loop is to ensure that we are not reusing type names and this is deliberately overriding it.

Alan-Cha:
I think this feature has the right intentions but in order to do it properly, fundamental changes need to be made.

"Used" type names are stored regardless of whether we intend to actually use them or not. See [here](https://github.com/IBM/openapi-to-graphql/blob/4f3d8cbce4224315ccaf22ea9b7087fd96e62825/packages/openapi-to-graphql/src/preprocessor.ts#L798). For example, in an API, there may be a person schema, for the person schema, we assign the `Person` type name as well as the `PersonInput` input type name, but the API may only require the `Person` type. Therefore, the `PersonInput` input type name has been taken up when we don't actually intend to use it.

Before adding the (input) type names to the `usedTypeNames` array, we should already have an idea of how that (input) type will be used in the context of the entire GraphQL API.

I think this whole ordeal is also loosely related to #274 which is also about improving the way we store names.

Alan-Cha:
However, I also wonder if this feature is too much of an edge case. Instead, I think there should be a generic option that allows you to define mappings from response/request bodies or schemas to names. This would be more useful in my opinion (and simpler to implement than the fundamental changes that I mentioned earlier) then a dedicated option focused only around request bodies with a name ending with "Input".

Alan-Cha:
To summarize, I like the `genericPayloadArgName` change. I also generally like the extended types change although I'd like to think about it some more. I think the "Input" name edge case handler will require some more work. 

What are you thoughts? And then we can discuss next steps

jun-sheaf:
> However, I also wonder if this feature is too much of an edge case. Instead, I think there should be a generic option that allows you to define mappings from response/request bodies or schemas to names.

This is a better solution. I thought of that as well, but in the end used the `simpleNames` solution since within the code, that option directly interacts with the `input`, so there would be some separation of concerns.

> I also generally like the extended types change although I'd like to think about it some more.

This is really minimal IMO. Considering we already used the JSON extended type (which doesn't make much sense since `null` is primitive in JSON so a `Void` type is more optimal) and every GraphQL I know uses extended types (in particular, the DateTime type), this change is necessary.

Alan-Cha:
Personally I'm in favor of breaking this up into two PRs. One that contains the `genericPayloadArgName` option as well as the extended types change and the other containing the WIP `Input` names change/generic schema name mapping change. What do you think?

jun-sheaf:
I agree. I just don't have the time right now, so it will take a while.

PR #364, opened 2021-02-25
check `.properties` for conflicts

### LINKED ISSUE #363: property named `type` is not included when aggregating field from an allOf
**Describe the bug**
When a JSONSchema `allOf` scope is reduced into a single GQL object type properties named `type`, `properties` or other JSONSchema keywords may not be included.

**To Reproduce**
Process an OpenAPI spec containing the following schema, the resulting FooBar GQL type will be missing the `type` field.

```yaml
Foo:
  type: object
  properties:
    value:
      type: string
    type:
      type: string  
Bar:
  type: object
  properties:
    baz:
      type: string
FooBar:
  type: object
  allOf:
     - $ref: '#/Foo'
     - $ref: '#/Bar'
```

**Expected behavior**
FooBar should include all properties of the composed sub schemas.

**Additional context**
This issue is due to the if statement here: 
https://github.com/IBM/openapi-to-graphql/blob/182f933a506ff0fd2db2b2418423041e898bd5d2/packages/openapi-to-graphql/src/preprocessor.ts#L1194-L1207
it should read `if (propertyName in collapsedSchema.properties)`

nlundquist:
Fixes #363


Alan-Cha:
Yes, this seem to be the right fix! Tomorrow, I will look into adding a test, as this clearly was not tested properly previously, and then accept it! Thank you once again, @nlundquist!

Alan-Cha:
I think the conflict resolution needs to be changed so that if there is a conflict, instead of simply ignoring the issue leading to unexpected behavior, we instead resolve to a [GraphQL JSON type](https://www.npmjs.com/package/graphql-type-json). This is a more involved change so I will file it as an issue. For the time being, I think we can accept this change.

Issue #357, opened 2021-01-29
Enums: keep original value option


PseudoSky:
**Describe the bug**
There isn't a way to keep the literal enum values from the oas spec.
The external api may require the enum to be exactly the same as what is defined.
Regardless of the options the values always seem to be sanitized to uppercase + underscore.

**Proposal**

- Map the enum values to valid values
- Store the mapping in state somehow
- Reverse the mapping of the values before the request is made

**OR**

I think it would be much more useful to add a fetch override option to the configuration -
so the consumer can implement more complex solutions / fix assumptions made by the library

- More easily map custom resolvers
- Add delegated resolvers
- Linking is currently near impossible with message envelopes
- Many more

No idea how complicated this would be within the library, I'm trying to avoid attempting it
Looking for thoughts on this

**To Reproduce**
Example
```yaml
TextFilter:
      description: Text filter
      required:
        - key
        - operator
        - value
      properties:
        key:
          type: string
          description: Key on an object
          enum:
            - id
            - name
            - assetType
        operator:
          type: string
          enum:
            - Equals
            - Does not equal
            - Contains
            - Does not contain
```

**Expected behavior**

I need the value posted to the external api to be "Does not equal" rather than "DOES_NOT_EQUAL"

Alan-Cha:
@PseudoSky Thanks for the issue and sorry for the late response. Could you verify that, indeed, the sanitzation/desanitization is not working? OtG is supposed to sanitize enum values to ALL_CAPS (to reflect GraphQL conventions), and desanitize back to the original values when the API call is made. I just want to make sure that this is not a misunderstanding.

If indeed, desanitization is functioning incorrectly, then this could potentially related to #274 where I discuss our naive approach to sanitization/desanitization mapping. However, if that is the case, I would assume that we would still see some kind of desanitization going on, just to an incorrect value.

However, I do agree, there probably should be an option that allows you to control whether the values are sanitized or not. I believe there are some strict limitations on the characters you can use in enum values but besides that, we can try to preserve as much of the original value as possible. 

***

> More easily map custom resolvers

We do offer custom resolvers via the `customResolvers` option. You can find that option [here](https://github.com/IBM/openapi-to-graphql/tree/master/packages/openapi-to-graphql#options). However, the enum values will be sanitized. If I recall correctly, we also expose the `sanitize()` and `desanitize()` functions. 

> Add delegated resolvers

[...]

PR #365, opened 2021-02-25
simple renaming also applies to schema values

nlundquist:
Possible fix for #357


Alan-Cha:
@nlundquist Thanks for your this PR as well!

I do wonder if we should consider having two different options, `simpleNames` and `simpleEnumValues`. I can see some limited use cases where you would want them separately but it is niche. 

If we want to combine it all under the same umbrella, I think we should consider changing the name of the option, perhaps `simpleSanitization`? Either way, this will be a breaking change.

Alan-Cha:
@nlundquist I think I would still like to keep field/type names separate from enums so I added the `simpleEnumValues` along with a test. Unless you have any other comments, I am going to accept this PR.

PR #372, opened 2021-03-08
Fix for issue #356

SimonCull:
Issue #356: https://github.com/IBM/openapi-to-graphql/issues/356

oas_3_tools - added 'strict' boolean argument to getValidOAS3 and used it to set laxDefaults in OASValidator.validate (ln 155)

index - passed options.strict to Oas3Tools.getValidOAS3

ran built & test

Signed-off-by: Simon Cull <simon.cull@schemeserve.com>

PR #375, opened 2021-03-16
Add validation options

Alan-Cha:
Replacement for #372


Alan-Cha:
There are also some assorted improvements to setting the default values of options.

PR #377, opened 2021-03-17
Add support for plain/text response type

### LINKED ISSUE #376: Cannot use plain/text as response type
**Describe the bug**
When using plain/text as a response type there is an error: TypeError: Cannot use 'in' operator to search for '$ref' in undefined

**To Reproduce**

Method `getResponseSchemaAndNames` for the following specification doesn't work properly.

```yaml
openapi: 3.0.3
info:
  version: v1
  title: Plain text example
paths:
  "/test":
    get:
      summary: Get string value
      responses:
        "200":
          description: Returns value as a string
          content:
            text/plain:
              example: example
```

**Expected behavior**
It should return string as a response type.

JKostov:
Fixes #376

### DIFF
--- a/packages/openapi-to-graphql/src/oas_3_tools.ts
+++ b/packages/openapi-to-graphql/src/oas_3_tools.ts
@@ -758,6 +758,22 @@ export function getResponseSchemaAndNames<TSource, TContext, TArgs>(
 
   if (responseContentType) {
     let responseSchema = responseObject.content[responseContentType].schema
+
+    if (!responseSchema && responseContentType === 'text/plain') {
+      return {
+        responseContentType,
+        statusCode,
+        responseSchemaNames: {
+          fromPath: inferResourceNameFromPath(path)
+        },
+        responseSchema: {
+          description:
+            'Placeholder to support operations with no response schema and text/plain response type',
+          type: 'string'
+        }
+      }
+    }
+
     let fromRef: string
     if ('$ref' in responseSchema) {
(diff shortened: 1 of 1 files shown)


Alan-Cha:
I need to take a closer look later but I believe [this](https://github.com/IBM/openapi-to-graphql/blob/784b89a64fae38ac8eb37ced6e66143056ae2afe/packages/openapi-to-graphql/src/oas_3_tools.ts#L781) is supposed to handle the case of non-application/json.

JKostov:
The problem is when there is no `schema` in the response type (like the example below) than the `responseSchema` is undefined on [this line](https://github.com/IBM/openapi-to-graphql/blob/784b89a64fae38ac8eb37ced6e66143056ae2afe/packages/openapi-to-graphql/src/oas_3_tools.ts#L766) and than we get the error from issue #376.

```yaml
paths:
  "/test":
    get:
      summary: Get string value
      responses:
        "200":
          description: Returns value as a string
          content:
            text/plain:
              example: example
```

Alan-Cha:
I see! In that case, could we just reorganize the code so that we still use [this](https://github.com/IBM/openapi-to-graphql/blob/784b89a64fae38ac8eb37ced6e66143056ae2afe/packages/openapi-to-graphql/src/oas_3_tools.ts#L781) but it checks if a `responseSchema` exists first?

JKostov:
Yes, it should fix the issue.

Alan-Cha:
@JKostov I've addressed this with #391. Thank you for all your help!

Issue #268, opened 2019-11-08
Feature request : provide custom Request module when generate resolver


moyara:
Hello, I have a feature request for custom request.
I see resolver_builder uses built in request module.
https://github.com/IBM/openapi-to-graphql/blob/master/packages/openapi-to-graphql/src/resolver_builder.ts#L15

It will be useful to provide custom request module such as custom-configured request or 3rd party packages like https://www.npmjs.com/package/retry-request

Alan-Cha:
@moyara I think it is possible to do this using the [`customResolvers`](https://github.com/IBM/openapi-to-graphql/tree/master/packages/openapi-to-graphql#options) option. Let me know if you think otherwise.

moyara:
> @moyara I think it is possible to do this using the [`customResolvers`](https://github.com/IBM/openapi-to-graphql/tree/master/packages/openapi-to-graphql#options) option. Let me know if you think otherwise.

But then I have to write All resolvers to apply custom request per each endpoint. I am lookinf for more general way to solve this without maintaining like 100 apis.

Alan-Cha:
@moyara I think currently we are not looking to support a number of different request modules. However, if `retry-request` has a similar API as `request` then we can consider adding it in, but this whole situation raises a number of questions concerning where does this decision lead to. @ErikWittern could you also weigh in?

ErikWittern:
For some larger context, OtG currently depends on the [request module](https://github.com/request/request) for invoking the REST API, which was a common option when we started the project. Since then, that module has actually [been put into maintenance mode only](https://github.com/request/request/issues/3142). So it may generally be a good thing to reconsider whether to use another option, ideally promise-based to allow for the convenient use of `async ... await`.

The specific question raised by this issue is whether users should be allowed to provide their own library instead of whichever one OtG depends on. I see two options to make that happen:

1. Providing a custom library is allowed, so long as it has the same API as the one used by OtG. E.g., replacing `request` by `request-retry` may easily be possible because they share (mostly) the same API (maybe this would be on the user's own risk of proper functionality).
2. OtG would need to provide an abstraction that allows to easily integrate any HTTP library, no matter what API it hast.

[...]

PR #381, opened 2021-03-22
fix: Added urlResolver on Resolver options

dinumathai:


Alan-Cha:
@dinumathai Thank you for this contribution! I generally like the idea of this PR. Giving full control over the requests that the resolvers make will make OtG more usable and gives more options to users.

However, I wonder if we should in a different way. I will give some background. In the beginning, there was only the `headers` and `qs` options for controlling requests. Later on, we added `requestOptions`. `requestOptions` provides more functionality than `headers` and `qs` so are thinking of deprecating `headers` and `qs`.

Another user also requested the programmatically control request headers so we modified `requestOptions` so it can accept a function for the `headers` field.

I do not want to add new request options (I would like to deprecate `headers` and `qs`) but after seeing your PR, I think it makes more sense that we allow the entire `requestOptions` to be a function, instead of just `headers`, and this way users can also specify the `url`.

Before:
```TypeScript
export type RequestHeadersFunction<TSource, TContext, TArgs> = (
  method: string,
  path: string,
  title: string,
  resolverParams?: {
    source: TSource
    args: TArgs
    context: TContext
    info: GraphQLResolveInfo
  }
) => Headers

export declare type RequestURLFunction<TSource, TContext, TArgs> = (method: string, path: string, title: string, resolverParams?: {
    source: TSource;
    args: TArgs;
    context: TContext;
    info: GraphQLResolveInfo;
}) => string;

export type RequestOptions<TSource, TContext, TArgs> = Omit<
  NodeRequest.OptionsWithUrl,
  'headers'
> & {
  headers?: Headers | RequestHeadersFunction<TSource, TContext, TArgs>
}

export type InternalOptions<TSource, TContext, TArgs> = {
  ...
  headers?: Headers | RequestHeadersFunction<TSource, TContext, TArgs>
  qs?: { [key: string]: string }
  urlResolver?: RequestURLFunction<TSource, TContext, TArgs>
  requestOptions?: Partial<RequestOptions<TSource, TContext, TArgs>>
  ...
}
```

After:
```TypeScript
export type RequestFunction<TSource, TContext, TArgs> = (
  method: string,
  path: string,
  title: string,
  resolverParams?: {
    source: TSource
    args: TArgs
    context: TContext
    info: GraphQLResolveInfo
  }
) => Partial<NodeRequest.OptionsWithUrl>

export type InternalOptions<TSource, TContext, TArgs> = {
  ...
  requestOptions: Partial<NodeRequest.OptionsWithUrl> | RequestFunction<TSource, TContext, TArgs>
  ...
}
```

Does this make sense? What do you think?

dinumathai:
That is a better solution. But do we need backward compatibility ?

Alan-Cha:
I am fine with making breaking changes as long as they are released in a major version and it is properly documented. However, I do have a few concerns about this change.

***

Currently, we perform an `Object.assign()` to the `requestOptions` (see [here](https://github.com/IBM/openapi-to-graphql/blob/784b89a64fae38ac8eb37ced6e66143056ae2afe/packages/openapi-to-graphql/src/resolver_builder.ts#L514)). If we want to give total control to the user, then we need to give a higher priority to `requestOptions` and even add the option to prevent OtG from changing it/for OtG to use the raw `requestOptions`. 

Perhaps we should also consider adding a new option, like `requestOptionsStrict` that is a boolean that controls this. If it is set to `false`, we stick to the original behavior, `Object.assign()`, and this is useful if a user just wants to use a new authentication header but does not want to change anything else. If it is set to `true`, then we use the raw `requestOptions` (the object or the output of the function), and this will give more confidence to users who want to have total control over everything. 

I'm concerned that there may be a third case. The `url` is also a field under `requestOptionsStrict`. I wonder if it may need special treatment. For example, there may be occasions where we want to strictly apply all of `requestOptions` but not `url` as well as occasions where we want to strictly apply all of `requestOptions`. Maybe we can make it so that if the user wants to strictly apply `requestOptions` and `url` is not provided, then we default to whatever OtG comes up with (as `url` is a required field).

I also wonder if we should add a new option like `requestOptionsStrict` or make `requestOptions` an object that also contains a `strict` field. I feel like the latter might be more intrusive to use. 

***

Currently, we are using the [request module](https://github.com/request/request) to perform our API calls. However, the module has been long deprecated and we need to use a new library. There is a PR for it #315 (as well as related #268). 

Until we find a solution for this, I hesitate to make very drastic changes. Therefore, I do not think we should remove `headers` and `qs` just yet. I still think that making `requestOptions` is still a useful feature until we finally switch libraries.

***

What do you think?

dinumathai:
Just wanted to put one more option for consideration. 

Shall we pass "options" as a parameter to the RequestOptionsFunction. This way developer will have more control. But we will need to validate for `url` and `method` in options.

```
export type RequestOptionsFunction<TSource, TContext, TArgs> = (
  options: Partial<NodeRequest.OptionsWithUrl>,
  method: string,
  path: string,
  title: string,
  resolverParams?: {
    source: TSource
    args: TArgs
    context: TContext
    info: GraphQLResolveInfo
  }
) => void
```

dinumathai:
@Alan-Cha, Added RequestOptionsFunction. Please review. 

Feel free to make any changes.

Alan-Cha:
> Shall we pass "options" as a parameter to the RequestOptionsFunction. 

Yes, this is a great idea. I was also thinking of this. 

***

@dinumathai I made a large number of changes, including restructuring the `resolver_builder.ts` to make it clearer and cleaner.

There is a complication that has come up that I wasn't aware of. This is a note that I have added to the `README.md` regarding the new `requestOptions`.

> If used as a function, this will be used to modify resolvers AND the schema. For example, if a `foo` header is set, then the resolver call will append the `foo` header, and the `foo` parameter will be removed from the GraphQL schema. In `resolver_builder.ts`, the function will be called with all arguments, but in `schema_builder.ts`, only the `method`, `path`, and `title` will be provided (because `resolverParams` and `generatedRequestOptions` are only available during execution). Therefore, for query, path, header, and cookie parameters, a value must be provided without `resolverParam` and `generatedRequestOptions` or else the schema will not have the right parameters.

First, let me know if this makes any sense to you. Secondly, I just want to point out that this has always been a problem, for example with using the `headers` option as a function. It was just never documented properly. Thirdly, I'm not very happy with this solution. This seems to force the user to do some unnatural things if he/she wants to use `resolverParams` and `generatedRequestOptions`.

Maybe there should be two different functions, one that is used for `resolver_builder.ts` and get all of the parameters and produces the request options, and one used for `schema_builder.ts` that gets only `method`, `path`, and `title` and returns true/false if certain arguments should be skipped in the schema.

I hope this all makes sense. If it's not intuitive, that is exactly why this is a big issue.

packages/openapi-to-graphql/lib/resolver_builder.d.ts
```diff
@@ -11,26 +11,31 @@ import { RequestOptions } from './types/options';
 import { GraphQLFieldResolver } from 'graphql';
 import { IncomingHttpHeaders } from 'http';
 export declare const OPENAPI_TO_GRAPHQL = "_openAPIToGraphQL";
-declare type GetResolverParams<TSource, TContext, TArgs> = {
+declare type GetSubscribeParams<TSource, TContext, TArgs> = {
```
Alan-Cha, line 14: Just cleaning up old code. Just ignore.

packages/openapi-to-graphql/lib/types/options.d.ts
```diff
@@ -97,7 +131,7 @@ export declare type InternalOptions<TSource, TContext, TArgs> = {
      * into a GraphQL ID type. To allow for more customzation, this option allows
      * users to specify other formats that should be interpreted as ID types.
      */
-    idFormats?: string[];
+    idFormats: string[];
```
Alan-Cha, line 134: More clean up.

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -559,92 +574,54 @@ export function getResolver<TSource, TContext, TArgs>({
...
-
-      // Add authentication cookie if created
-      if (authCookie !== null) {
-        const j = NodeRequest.jar()
-        j.setCookie(authCookie, options.url)
-        options.jar = j
-      }
-    }
+    // The options `requestOptions` will finalize the options
+    if (requestOptions) {
```
Alan-Cha, line 610: `requestOptions` has been moved to the very bottom, right before the call. This allows `requestOptions` to have the final say on how the request should be made.

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -559,92 +574,60 @@ export function getResolver<TSource, TContext, TArgs>({
...
+       * use the original value
+       */
+      const requestOptionsValue =
+        typeof requestOptions === 'function'
+          ? requestOptions(
+              method,
+              path,
+              title,
+              { source, args, context, info },
+              options
```
Alan-Cha, line 622: This is where we also provide the generated Request options as you suggested.

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -559,92 +574,60 @@ export function getResolver<TSource, TContext, TArgs>({
...
+          : requestOptions
 
-    // Extract OAuth token from context (if available)
-    if (data.options.sendOAuthTokenInQuery) {
-      const oauthQueryObj = createOAuthQS(data, context)
-      Object.assign(options.qs, oauthQueryObj)
-    } else {
-      const oauthHeader = createOAuthHeader(data, context)
-      Object.assign(options.headers, oauthHeader)
+      options = merge(options, requestOptionsValue)
```
Alan-Cha, line 626: Notice that we merge the original options with `requestOptionsValue`. If we were to implement a `requestOptionsStrict` option like I mentioned previously, if would do something here.

packages/openapi-to-graphql/src/schema_builder.ts
```diff
@@ -1026,80 +1026,63 @@ function linkOpRefToOpId<TSource, TContext, TArgs>({
 }
 
 /**
- * Determin if an argument should be created if the argument has already been
+ * Determine if an argument should be created if the argument has already been
  * provided through the options
  */
-function skipArg<TSource, TContext, TArgs>(
+function shouldSkipArgumentBecauseOptions<TSource, TContext, TArgs>(
```
Alan-Cha, line 1032: This is where the problem I mentioned before comes from.

`getResolver` in `resolver_builder.ts` uses `requestOptions` to change the API call.

`shouldSkipArgumentBecauseOptions` in `schema_builder.ts` uses `requestOptions` to determine if it should add an argument to the field in the GraphQL

They both provide different arguments.

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -559,92 +574,60 @@ export function getResolver<TSource, TContext, TArgs>({
...
-    }
+    // The options `requestOptions` will finalize the options
+    if (requestOptions) {
+      /**
+       * If requestOptions is a function, the run the function, otherwise,
+       * use the original value
+       */
+      const requestOptionsValue =
+        typeof requestOptions === 'function'
+          ? requestOptions(
```
Alan-Cha, line 617: Here we provide `method`, `path`, `title`, resolver arguments, and generated options

packages/openapi-to-graphql/src/schema_builder.ts
```diff
@@ -1026,80 +1026,63 @@ function linkOpRefToOpId<TSource, TContext, TArgs>({
...
-
-          if (typeof headers === 'object') {
-            return true
-          }
+  // Get the value of `requestOptions` as it can be both an object or a function
+  const requestOptionsValue =
+    typeof data.options.requestOptions === 'object'
+      ? data.options.requestOptions
+      : typeof data.options.requestOptions === 'function'
+      ? data.options.requestOptions(
```
Alan-Cha, line 1042: Here we only provide `method`, `path`, and `title`. No resolver arguments or generated request options.

packages/openapi-to-graphql/test/example_api.test.ts
```diff
@@ -2274,18 +2272,17 @@ test('Use headers option as function', () => {
...
 })
 
-test('Use requestOptions headers option as function', () => {
+test('Use requestOptions option as function', () => {
   const options: Options<any, any, any> = {
-    requestOptions: {
-      headers: (method, path, title) => {
-        if (method === 'get' && path === '/snack') {
-          return {
+    requestOptions: (method, path, title) => {
```
Alan-Cha, line 2277: Example of using requestOptions as a function.

dinumathai: Looks good

dinumathai:
Planned any date for merge ?

Alan-Cha:
@dinumathai I have a lot of concerns about whether this is the right design. As I mentioned previously, `requestOptions` is used by both the `resolver_builder.ts` and `schema_builder.ts` and they provide different arguments. Do you have any opinions on this?

dinumathai:
> @dinumathai I have a lot of concerns about whether this is the right design. As I mentioned previously, `requestOptions` is used by both the `resolver_builder.ts` and `schema_builder.ts` and they provide different arguments. Do you have any opinions on this?

1. Let `requestOptions` be only an object. Add another one like `requestOptionsFunc` which can be only function.
2. Update README - explaining about this.

dinumathai:
Or shall we move back to initial option of adding new parameter ?

PR #383, opened 2021-03-29
Add support for processing discriminator property used with allOf property

mladjanm:
This pull request introduces support for handling polymorphic schemas in the Open API 3 Specification responses that use the discriminator property with the allOf property to describe polymorphism. 

Here is an example situation that this pull request resolves:
```yaml
components:
  schemas:
    Pet:
      type: object
      discriminator:
        propertyName: petType
      properties:
        name:
          type: string
        petType:
          type: string
          enum:
          - Cat
          - Dog
      required:
      - name
      - petType
    Cat:
      description: A representation of a cat
      allOf:
      - $ref: '#/components/schemas/Pet'
      - type: object
        properties:
          huntingSkill:
            type: string
            description: The measured skill for hunting
            enum:
            - clueless
            - lazy
            - adventurous
            - aggressive
        required:
        - huntingSkill
    Dog:
      description: A representation of a dog
      allOf:
      - $ref: '#/components/schemas/Pet'
      - type: object
        properties:
          packSize:
            type: integer
            format: int32
            description: the size of the pack the dog is from
            default: 0
            minimum: 0
        required:
        - packSize
```

In this example an union called `Pet` would be created and it would be composed of two types: `Dog` and `Cat`.


Alan-Cha:
@mladjanm Wow! This seems like an awesome change! Thank you so much for this PR! I will take a look at this and try to get it in asap.

Warxcell:
:+1:

vigie:
Wouldn't a more natural GQL implementation of this be to turn `Pet` into an `interface` and have `Cat` and `Dog` be two `types ` that `implement Pet`?

By using union, we're losing some of the information that `allOf` is giving us.

TrivialNinja:
Do we have any plan on having this support?

Alan-Cha:
@vigie Yes, on the surface level, that makes sense to me. We just have not implemented this kind of feature. I suppose if all member schemas reference one or more other schemas, then we can create interfaces from all those referenced schemas. I just wonder if there is any kind overlap between union and interface that we need to keep in mind.

@TrivialNinja I’m planning on reviewing new and old PRs again this week.

PR #394, opened 2021-04-22
Update anyOf and oneOf

Alan-Cha:


Alan-Cha:
This is a breaking change. 

I had to change the `anyOf test with no object type member schemas\n\nEquivalent to GET /anyOf8` test in `packages/openapi-to-graphql/test/example_api4.test.ts`.

Alan-Cha:
To summarize, this introduces the `TargetGraphQLType` enum and moves some of the complex logic resolving `anyOf` and `oneOf` from `createDataDef()` into `getTargetGraphQLType()`. We have a switch statement in `createDataDef` that decides how the `dataDefinition` should be created, based on the `targetGraphQLType`. Previously, we interjected code that would handle `anyOf` and `oneOf` before the switch statement. Now, we check the `anyOf` and `oneOf` in `getTargetGraphQLType` and use the switch statement as all other cases are handled.

PR #407, opened 2021-06-28
Replace deprecated request with cross-fetch

ardatan:
Replace deprecated `request` lib with w3 compatible `cross-fetch` library
`fetch` option added to provide custom fetch impl


packages/openapi-to-graphql/src/auth_builder.ts
```diff
@@ -244,7 +247,8 @@ function getViewerOT<TSource, TContext, TArgs>(
 function getViewerAnyAuthOT<TSource, TContext, TArgs>(
   name: string,
   queryFields: GraphQLFieldConfigMap<any, any>,
-  data: PreprocessingData<TSource, TContext, TArgs>
+  data: PreprocessingData<TSource, TContext, TArgs>,
+  fetch: typeof global['fetch']
```
Alan-Cha, line 251: Why is this `typeof global['fetch']` instead of `typeof crossFetch`?

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -629,253 +620,278 @@ export function getResolver<TSource, TContext, TArgs>({
...
+
+      Object.assign(qs, oauthQueryObj)
     } else {
       const oauthHeader = createOAuthHeader(data, context)
       Object.assign(options.headers, oauthHeader)
     }
 
     resolveData.usedRequestOptions = options
     resolveData.usedStatusCode = operation.statusCode
+    const setSearchParams = (obj: any, path: string[]) => {
```
Alan-Cha, line 644: Why not pull this out?

packages/openapi-to-graphql/src/auth_builder.ts
ardatan: Fixed :)

packages/openapi-to-graphql/src/resolver_builder.ts
ardatan: Yes you're right :) Done!

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -13,18 +13,17 @@ import { ConnectOptions } from './types/options'
...
 // Imports:
 import * as Oas3Tools from './oas_3_tools'
-import * as querystring from 'querystring'
 import * as JSONPath from 'jsonpath-plus'
 import { debug } from 'debug'
 import { GraphQLError, GraphQLFieldResolver } from 'graphql'
 import formurlencoded from 'form-urlencoded'
 import { PubSub } from 'graphql-subscriptions'
-import { IncomingHttpHeaders } from 'http'
+import urljoin = require('url-join')
```
Alan-Cha, line 26: Can we avoid the `require`?

ardatan: Bob can handle it I guess

Alan-Cha:
See #413

PR #408, opened 2021-06-28
Use GraphQL Scalars

ardatan:
- Use GraphQLBigInt for `int64`
- Use GraphQLJSON instead of `graphql-type-json`
- - We could still use graphql-type-json but since graphql-scalars has already `GraphQLJSON`, I just wanted to use one library for both JSON and BigInt scalars.


Alan-Cha (approved):
Looks good to me!

Alan-Cha:
See #415

PR #409, opened 2021-06-28
Export translateOpenAPIToGraphQL

ardatan:
Export translateOpenAPIToGraphQL to let users bypass OAS validation and swagger openapi translation which will help reduce the bundle size if people want to exclude oas validator and swagger converter libs in the final bundle


Alan-Cha (approved):

Alan-Cha:
Great change!

PR #434, opened 2021-12-02
Add support for Graphql file upload

eokoneyo:
Hi There, 

Really great work with what's been done so far with the implementation to translate OAS to GraphQL, I recently had some reason whilst using this library to support file uploads for the GraphQL server running on the schema generated by the library. I know there's no support for it and it's already been brought up as seen here https://github.com/IBM/openapi-to-graphql/issues/220

So I added support for file uploads, Here are the considerations I made;
- API definitions with content-typeof `multipart/form-data` are processed as such
- the implementation for file upload is solely based on the spec proposal defined here https://github.com/jaydenseric/graphql-multipart-request-spec which would imply that users who wish to use this need to make some extra configuration for parsing request body to match the spec's expectations.
- Uploaded files are streamed directly to the intending REST endpoint, this way the library needs not necessarily concern itself too much about memory management and the problems that could accompany it when multiple users are uploading files at scale. 

I would appreciate feedback pertaining to this and like to see this merged in, as I'm using the changes in here as a patch to the library for upload support in my team.

PR #435, opened 2021-12-15
Refactor #434

Alan-Cha:


eokoneyo:
@Alan-Cha I think something went wrong with the release, there's a new version on yarn and npm quite alright but it doesn't contain the changeset from this PR

PR #440, opened 2022-01-23
Extend OAS implementations

eokoneyo:
Hi, following our discussion in https://github.com/IBM/openapi-to-graphql/pull/426 , I decided to open a different PR to tackle couple of issues I found pertaining to how the definitions from an OAS might be processed.

Here's a couple of issues the PR addresses; 

- Fixes implementation where the `style` and `explode` properties for endpoint parameters didn't adhere to the Open API Spec. Given the partial definition;
    ```
    "/api/path": {
      get: {
       "parameters" : [ {
              "name" : "parameters",
              "in" : "query",
              "required" : true,
              "style" : "form",
              "explode" : true,
              "schema" : {
                "$ref" : "#/components/schemas/SomeSchemaInput"
              }
        } ],
      }
    }
    ```
    where `SomeSchemaInput` has properties of `filter` and `limit` for example, I found that the http request made for the generated GraphQL resolver is of the format; `http://{baseUrl}/api/path?parameters[limit]=10&parameters[offset]=10` when the http request url should rather be `http://{baseUrl}/api/path?limit=10&offset=10` considering the provided values for style and explode, as defined in Open API specification. Relates to https://github.com/IBM/openapi-to-graphql/pull/312 but also includes support for Arrays and Objects
- Also fix issue where an error is thrown when link is created with parameter that's a nested object; for instance given the link definition;
    
    ```
    ...
            responses: {
              200: {
                links: {
                  invitations: {
                    operationId: 'someOperationId',
                    parameters: {
                      filter: {
                        entityId: '$response.body#/id',
                      },
                    },
                  },
                },
              },
    
    ```
    previously filter would be attempted to be processed as a string, which resulted in errors.

### DIFF
--- a/packages/openapi-to-graphql/src/resolver_builder.ts
+++ b/packages/openapi-to-graphql/src/resolver_builder.ts
@@ -1360,7 +1360,23 @@ export function extractRequestDataFromArgs<TSource, TContext, TArgs>(
 
         // Query parameters
         case 'query':
-          qs[param.name] = args[saneParamName]
+          // setting param style as form assumes explode is true by default
+          if (param.style === 'form' && Object.prototype.toString.call(args[saneParamName]) === '[object Object]') {
+            if (param.explode === false) {
+              qs[param.name] = Object.entries(args[saneParamName]).reduce((acc, val) => {
+                acc += val.join(',')
+                return acc
+              }, '')
+            } else {
+              Object.entries(args[saneParamName]).forEach(([key, value]) => {
+                qs[key] = value
+              })
+            }
+          } else if (Array.isArray(args[saneParamName]) && param.style === 'form' && param.explode !== false) {
+            qs[param.name] = args[saneParamName].join(',')
+          } else {
+            qs[param.name] = args[saneParamName]
+          }
           break
 
         // Header parameters
--- a/packages/openapi-to-graphql/src/resolver_builder.ts
+++ b/packages/openapi-to-graphql/src/resolver_builder.ts
@@ -55,7 +55,7 @@ type AuthOptions = {
 
 type GetResolverParams<TSource, TContext, TArgs> = {
   operation: Operation
-  argsFromLink?: { [key: string]: string }
+  argsFromLink?: { [key: string]: any }
   payloadName?: string
   responseName?: string
   data: PreprocessingData<TSource, TContext, TArgs>
@@ -63,6 +63,14 @@ type GetResolverParams<TSource, TContext, TArgs> = {
   requestOptions?: Partial<RequestOptions<TSource, TContext, TArgs>>
   fileUploadOptions?: FileUploadOptions
   fetch: typeof crossFetch
+}
+
+type inferLinkArgumentsParam<TSource, TContext, TArgs> = {
+  paramName: string
+  value: any
+  resolveData: Partial<ResolveData<TSource, TContext, TArgs>>
+  source: TSource
+  args: TArgs
 }
 
 type GetSubscribeParams<TSource, TContext, TArgs> = {
@@ -325,7 +333,7 @@ export function getPublishResolver<TSource, TContext, TArgs>({
 }
 
 /**
- *  Returns values for link arguments, also covers the cases for
+ * Returns values for link arguments, also covers the cases for
  * if the link parameter contains constants that are appended to the link parameter
  *
  * e.g. instead of:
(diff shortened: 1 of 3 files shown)


Alan-Cha:
@eokoneyo Thank you so much for the follow up! I will take a closer look at this later today and try to get it in. This is looking great so far!

Alan-Cha:
Okay, I have reviewed this and I generally like what I see! I have a few comments, however.

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -324,6 +332,46 @@ export function getPublishResolver<TSource, TContext, TArgs>({
...
+ * abc_{$response.body#/employerId}
+ */
+function inferLinkArguments<TSource, TContext, TArgs>({
+  paramName,
+  value,
+  resolveData,
+  source,
+  args
+}: inferLinkArgumentsParam<TSource, TContext, TArgs>)  {
+  if (Object.prototype.toString.call(value) === '[object Object]') {
```
Alan-Cha, line 352: I don't really like this pattern. It seems a bit lengthy and we don't use this pattern anywhere else in our code base. I believe that `typeof value === 'object'` should be sufficient for this use case. Feel free to correct me if I am wrong.

packages/openapi-to-graphql/test/example_api.test.ts
```diff
@@ -2411,3 +2411,32 @@ test('Non-nullable properties from nested allOf', () => {
...
+          id
+        }
+      }
+    }
+  `
+
+  await graphql(createdSchema, query).then((result) => {
+    // target error field because the corresponding server url is not implemented,
+    // also we get the full request url as in failed request errors
+    result.errors.forEach((error) => {
```
Alan-Cha, line 2433: Thank you so much for adding a test case! Adding test cases make my life so much easier.

This is certainly an interesting way to test this functionality, to create an error and examine the request. I think it would be really interesting if we could extract the backend requests that we are making and use them for test cases but I think this pattern isn't the best way to do this.

I am currently working on another issue that I have come across which will also require me to rework this test. We can accept this for now.

packages/openapi-to-graphql/src/resolver_builder.ts
eokoneyo: using the prototype method might be overkill, but it does guarantee we are dealing with a real object but I think `typeof` should suffice in this case.

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -324,6 +324,38 @@ export function getPublishResolver<TSource, TContext, TArgs>({
...
+ *
+ * e.g. instead of:
+ * $response.body#/employerId
+ *
+ * it could be:
+ * abc_{$response.body#/employerId}
+ */
+function inferLinkArguments (paramName, value, resolveData, source, args) {
+  if (Object.prototype.toString.call(value) === '[object Object]') {
+    return Object.entries(value).reduce((acc, [key, value]) => {
```
Alan-Cha, line 353: Thanks for fixing this issue as well! However, I think we are missing a test for this and while I was trying to add it in, I came across some other issues...

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -324,6 +324,38 @@ export function getPublishResolver<TSource, TContext, TArgs>({
...
+      acc[key] = inferLinkArguments(paramName, value, resolveData, source, args)
+      return acc
+    }, {})
+  }
+
+  if (value.search(/{|}/) === -1) {
```
Alan-Cha: ... namely, we cannot assume that `value` is a string. There seems to be a mismatch with our type definitions and OAS.

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -55,7 +55,7 @@ type AuthOptions = {
 
 type GetResolverParams<TSource, TContext, TArgs> = {
   operation: Operation
-  argsFromLink?: { [key: string]: string }
+  argsFromLink?: { [key: string]: any }
```
Alan-Cha, line 58: So I had to make a change here.

Source: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#link-object

packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -324,6 +332,46 @@ export function getPublishResolver<TSource, TContext, TArgs>({
...
+  args
+}: inferLinkArgumentsParam<TSource, TContext, TArgs>)  {
+  if (typeof value === 'object') {
+    return Object.entries(value).reduce((acc, [key, value]) => {
+      acc[key] = inferLinkArguments({paramName, value, resolveData, source, args})
+      return acc
+    }, {})
+  }
+
+  if (typeof value !== 'string') {
```
Alan-Cha, line 359: As well as a check here. However, I am still working on finishing a test but unfortunately I don't think I will be able to finish this today and I really want to get this checked in. I think for our purposes, this PR is good enough to get merged in!

Alan-Cha:
@eokoneyo It's in! Thanks so much again! I know this fix and the old PR has been getting really stale. This was a really great change! Thank you once again!

packages/openapi-to-graphql/src/resolver_builder.ts
eokoneyo: right... I totally missed this, thanks for spotting this

PR #444, opened 2022-02-03
Non-string and nested link object parameters

Alan-Cha:


packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -1225,45 +1226,37 @@ function resolveRuntimeExpression(
...
-        path: value.split('body#/')[1],
-        json: root
-      })
-      if (Array.isArray(tokens) && tokens.length > 0) {
-        return tokens[0]
-      } else {
-        httpLog(`Warning: could not extract parameter '${paramName}' from link`)
-      }
+    } else if (runtimeExpression.startsWith('$response.body#')) {
+      return JSONPointer.get(root, runtimeExpression.split('body#')[1])
```
Alan-Cha, line 1230: Switch from JSONPath to JSONPointer. It did not make sense to use JSONPath here. It just happened to work for simple parameters.

Alan-Cha:
Related to conversation in #440

PR #449, opened 2022-03-04
check operation responseContentType before checking if it includes a content type

### LINKED ISSUE #448: fillEmptyResponses does not work for 204 responses that return a content type
**Describe the bug**
Unable to use an OpenAPI endpoint that has no return type.

**To Reproduce**
1. Define a swagger schema that contains a schema-less response, 
2. Server has a schema-less endpoint that responds with a 204 and application/json as the content type
3. Use `createGraphQLSchema` to export a graphql executor with the option `fillEmptyResponses: true`
4. Execute a query or mutation against endpoint
5. Observe that the operation gets executed but that the endpoint returns with an error.

**Expected behavior**

Either for the operation to return successfully with text or undefined return type OR for `createGraphQLSchema` to fail early if `responseContentType` is undefined

**Screenshots**

Server will respond with:
```
{
    "stacktrace": [
        "TypeError: Cannot read property 'includes' of undefined",
        "    at /code/node_modules/dist/openapi-to-graphql/src/resolver_builder.js:472:38",
        "    at processTicksAndRejections (internal/process/task_queues.js:95:5)"
    ]
}
```

If it's swagger schema looks like:
```
"put": {
        "description": "Update a building type",
        "operationId": "update_building_type",
        "parameters": [
          {
            "in": "body",
            "name": "building_type",
            "schema": {
              "$ref": "#/definitions/BuildingType"
            }
          }
        ],
        "responses": {
          "204": {
            "description": "Success."
          },
          "401": {
            "$ref": "#/responses/UnauthorizedError"
          },
          "403": {
            "$ref": "#/responses/ForbiddenError"
          },
          "default": {
            "$ref": "#/responses/GenericError"
          }
        },
```

**Additional context**
From what I've seen this happens when we call `responseContentType.includes` and `responseContentType` is undefined. This appears to happen when the source schema does not define a ref or schema.

zbyte64:
Some API implementations will respond with a content type header for a `204 NO_CONTENT` response. The existence of that header allows for the code to call includes on a variable that is possibly undefined.

* check operation responseContentType before trying to check if it includes a content type

Fixes #448

### DIFF
--- a/packages/openapi-to-graphql/src/resolver_builder.ts
+++ b/packages/openapi-to-graphql/src/resolver_builder.ts
@@ -772,7 +772,7 @@ export function getResolver<TSource, TContext, TArgs>({
     } else {
       httpLog(`${response.status} - ${Oas3Tools.trim(body, 100)}`)
 
-      if (response.headers.get('content-type')) {
+      if (response.headers.get('content-type') && operation.responseContentType) {
         /**
          * Throw warning if the non-application/json content does not
          * match the OAS.


Alan-Cha:
Thank you so much for this fix! It looks good to me! Could you just sign off you commit and then I will accept it and make a new release.

Should be something along the lines of

```
git commit --amend -s
git push -f
```

zbyte64:
> Thank you so much for this fix! It looks good to me! Could you just sign off you commit and then I will accept it and make a new release.
> 
> Should be something along the lines of
> 
> ```
> git commit --amend -s
> git push -f
> ```
@Alan-Cha 
Done, thank you for the simple instructions.

spinningarrow:
@Alan-Cha any idea when this will be released? Ran into this issue and I see the last release was in February, before this change was merged to master.

Issue #421, opened 2021-09-09
Arrays in query strings with cross-fetch


thomsonbw:
"request" used the "qs" query string manipulation package and exposed its options in the requestOptions object.  One of these, the "arrayFormat" setting,  supported a variety of formats for an array parameter { a: [ "b", "c"] } 
    'indices'        a[0]=b&a[1]=c    (the default)
    'brackets'      a[]=b&a[]=c
    'repeat'         a=b&a=c
    'comma'        a=b,c

With the switch to cross-fetch this flexibility was lost, as query parameters are handled by resolver_builder code that only implements the first format.

eokoneyo:
@thomsonbw I've also stumbled on this problem, the unfortunate issue is there is no standard way of handling arrays in query strings, there's a whole discussion about this here https://stackoverflow.com/questions/6243051/how-to-pass-an-array-within-a-query-string. I wonder if it might make sense to have a config that allows the end user to specify how they'd like their arrays processed

thomsonbw:
@eokoneyo we effectively had a config option before cross-fetch.  You could set the arrayFormat in the o2g requestOptions object and it would be passed through to the underlying package, but the current code doesn't provide similar flexibility.

Alan-Cha:
@thomsonbw Sorry for the very late reply. I'm also sorry that this functionality was lost. I'll try to look into this but if any of you have suggestions on how this could be resolved, I'd love to know.

DanailPenev:
@thomsonbw, I have opened a Pull request that adds an option for one of the other formats https://github.com/IBM/openapi-to-graphql/pull/453

nicolasparada:
It's true that there are many ways to define arrays in query strings, but `URLSearchParams` API uses the "repeat" format:

```js
const q = new URLSearchParams()
q.append("test", "one")
q.append("test", "two")
q.append("test", "three")
q.toString() === "test=one&test=two&test=three"
```

It should at the very least support that.

PR #453, opened 2022-04-02
Add optional query string usage in resolvers

DanailPenev:
# Summary
- Add query-string library as a dependency
- Add an optional parameter `useQueryString` to `RequestOptions` which allows callers to use the query string library when generating the resolvers
- This enables use cases such as sending array query parameters like `?x=a&x=b&x=c` compared to the existing behaviour of `?x[0]=a&x[1]=b&x[2]=c`
- The query string library does not behave nicely with nested objects, such as the test case in `packages/openapi-to-graphql/test/example_api6.test.ts` so if we want to fully migrate to using it, we might have to add some custom logic

# Motivation
- This resolves https://github.com/IBM/openapi-to-graphql/issues/421 

# Testing
- Add a unit test that verifies the behaviour with or without the new optional parameter
- Verify it's working as intended in a private project


thomsonbw:
Thanks, Danail.  We have a private fix too, but instead of a boolean option we added a function-valued option "queryStringify" so the application could implement whatever strategy it wanted.  I was reluctant to propose it here though because I didn't think a global setting is the correct answer when graphql is aggregating REST services from several sources.  Using Openapi's style/format/explode properties seems better, since they allow different services to use different strategies, but I'm not sure they cover all the array formats we are likely to run into.  What is your experience?

DanailPenev:
@thomsonbw, I agree with you completely that the change I am proposing is not a perfect solution. Ideally, the parameter serialization should be driven by the OpenAPI spec to conform to the different array query parameter formats there are with an optional override if something is going wrong.

I think the boolean option is a good middle ground right now because it exposes a straightforward solution to others who might be struggling with the same issue. It also restores the previously existing functionality that was removed after 2.4.0 when the request library was swapped for `cross-fetch`.

PR #454, opened 2022-04-03
Allow excepting keys from sanitization

DanailPenev:
# Summary
- Add an optional parameter nonSanitizableObjectKeys that prevents OpenAPI-to-GraphQL from sanitizing objects that fall under these keys
- This does not conform to the GraphQL rules but allows power users who have a non-standard schema to persevere their original data in exceptional cases
- I understand this might be controversial so happy to explore alternative options that are not much larger in scope

# Testing
- Add a test case for the sanitization function
- Add an E2E test case for the server and schema creation with appropriate params
- All other tests pass

### DIFF
--- a/packages/openapi-to-graphql/src/index.ts
+++ b/packages/openapi-to-graphql/src/index.ts
@@ -104,6 +104,7 @@ const DEFAULT_OPTIONS: InternalOptions<any, any, any> = {
   genericPayloadArgName: false,
   simpleNames: false,
   simpleEnumValues: false,
+  nonSanitizableObjectKeys: [],
   singularNames: false,
   createSubscriptionsFromCallbacks: false,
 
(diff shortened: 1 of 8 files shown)


Alan-Cha:
@DanailPenev Thanks for providing this suggestion and including tests! It is also so helpful to include tests.

However, I wonder if there is a better way to do this. Are you aware of the [custom type/field/enum extension](https://github.com/IBM/openapi-to-graphql/tree/master/packages/openapi-to-graphql#custom-type-and-field-names-and-enum-values)? The intention of this extension is to allow users to have greater control over type and field names and enum values. However, I think we still ensure basic santization aka remove GraphQL unsafe characters. 

Perhaps there should be an option that prevents OtG from sanitizing these extension names/values giving users definitive control over names and values. Perhaps something like `disableCustomNameValueSanitzation`? It's a long name so I am open to alternatives.

In general, what do you think?

Alan-Cha:
And I am sorry for the late reply. I have really been struggling to maintain OtG by myself.

PR #475, opened 2022-09-02
Fix queries with array parameters

### LINKED ISSUE #474: Queries with array parameters doesn't work
**Describe the bug**
Queries with query parameters that are array based doesn't work.

It boils down to this line:
https://github.com/IBM/openapi-to-graphql/blob/89515187ed664370a9bbc1cd58f843412d20741c/packages/openapi-to-graphql/src/resolver_builder.ts#L1374

An array is also considered an object in JavaScript. The if statement needs to be flipped.

Zn4rK:
Fixes #474


Zn4rK:
Would be super pleased if this could be merged and released @Alan-Cha. I wrote test cases to back up this work and fix the issue mentioned in the PR.

BradSiegler:
Just bumping the importance of this for my project.

PR #471, opened 2022-07-18
Add support for `graphql@16`

mhassan1:
This PR adds support for `graphql@16`. This should be part of a semver major release.

Resolves https://github.com/IBM/openapi-to-graphql/issues/458.

PR #482, opened 2023-02-28
Hotfix #471

Alan-Cha:
I preemptively merged #471 without letting the CI run and it turns out, many things started breaking. This is an attempt to address those issues.


packages/openapi-to-graphql/src/resolver_builder.ts
```diff
@@ -904,7 +904,7 @@ export function getResolver<TSource, TContext, TArgs>({
               let arraySaneData = saneData
 
               if ('limit' in args) {
```
Alan-Cha, line 906: All of these `extends object` changes come from this line. Otherwise, TS complains that I cannot access `limit` in `args`.

Alan-Cha: I am not sure if this is the best approach. Please let me know if anyone else has any other opinions.

Alan-Cha:
The current concerns are the example API 7 test, the one for subscriptions, and a type mismatch [here](https://github.com/Alan-Cha/oasgraph/blob/0ddf0334dd07a70cd1b33df28ee00ae8e4d81e07/packages/openapi-to-graphql-cli/src/index.ts#L244) between the CLI and the main library.

packages/openapi-to-graphql/src/resolver_builder.ts
bitiz: This seems OK to me, also good to have TArgs a it more restrictive now.

I'm not sure about this though, as limit could potentially be a string (or other type). Are the values sanitized?
```
const limit = args['limit'] as number
```

Pagebakers: This seems OK to me, also good to have TArgs a it more restrictive now.

I'm not sure about this though, as limit could potentially be a string (or other type). Are the values sanitized?
```
const limit = args['limit'] as number
```

Pagebakers:
> here

The package that is used in the test is actually archived and not maintained anymore, should migrate to use:
https://github.com/enisdenjo/graphql-ws

Pagebakers:
This fixes the test, not all required props are passed down.

```ts
        subscriptionServer = new SubscriptionServer(
          {
            execute: (
              schema,
              document,
              rootValue,
              contextValue,
              variableValues,
              operationName
            ) =>
              execute({
                schema,
                document,
                rootValue,
                contextValue,
                variableValues,
                operationName
              }),
            subscribe: (
              schema,
              document,
              rootValue,
              contextValue,
              variableValues,
              operationName
            ) => {
              return subscribe({
                schema,
                document,
                rootValue,
                contextValue,
                variableValues,
                operationName
              })
            },
            schema,
            onConnect: (params, socket, context) => {
              // Add pubsub to subscribe context
              return { pubsub }
            }
          },
          {
            server: wsServer,
            path: '/subscriptions'
          }
        )
```

packages/openapi-to-graphql/src/resolver_builder.ts
Alan-Cha: Good point, I added a type check

Alan-Cha:
@Pagebakers Thanks for your suggestions! I'll make a new release now. We definitely need to upgrade the other modules but I don't want to delay things any further.

Pagebakers:
Nice! I will see if there are other low hanging fruits this week.

PR #490, opened 2023-03-01
fix: graphql-upload types

Pagebakers:
@types/graphql-upload wasn't updated and causing issues with consuming the library.

We should enable ts strict mode to better pick up on these issues.


Alan-Cha:
Rebased and accepted in #491

